PR #40, opened 2013-12-18
Added BDD tests for eZ Demo Blog Page

pedroresende:


andrerom:

bdunogier:
Hmmm, I'm actually re-introducing demo with content...

andrerom:
Yes, but I still would propose we change the BDD test approach over time, as the content keeps on changing meaning PR's like these are left behind (this one is one year old, others where older).

PR #41, opened 2013-12-20
This function was moved to here due to the BrowserContext re-factoring

pedroresende:
When re-factoring BrowserContext.php the function has been moved.
https://github.com/ezsystems/ezpublish-community/pull/90


andrerom:
@ezsystems/qa-team Closing this as demo bundle will be w/o change password page now that legacy is out, and when we re implement it we can re visit this topic.

PR #53, opened 2014-02-28
BDD DemoContent: Added implementations for home page side bar

jpingo:
Created Implementation file for specific Feature testing on DemoContent
Added common paths for pages of DemoContent Installation


andrerom:

PR #54, opened 2014-02-28
Added BDD Tests for QA-275 As QA I want to do BDD tests for Start page / side bar

jpingo:
Created feature file for testing the Home Page Side Bar


andrerom:

PR #69, opened 2014-04-19
BDD: scen. for shown/hidden website toolbar

bdunogier:
> Status: review
> Uses https://github.com/bdunogier/ezpublish-community/tree/demobundle-bdd/website_toolbar.

Tests that the website toolbar doesn't show up for anonymous users, and does for admin.


andrerom:
Closing as we just removed this, however would make sense to backport and cover preview as well as it should be hidden there.

PR #130, opened 2015-01-20
Sync .less and .css files after fixing EZP-23829

dpobel:
Same as in https://github.com/ezsystems/ezdemo/pull/40 (even if because of https://jira.ez.no/browse/EZP-23919 it's not really needed)


lolautruche:
+1

yannickroger:
+1

PR #131, opened 2015-01-21
EZP-23513: render embed link if defined

pspanja:
This PR resolves issue https://jira.ez.no/browse/EZP-23513

This is a missing piece on the new stack part of the issue, adding link rendering if embed is linked.


pspanja:
Does anyone knows about the failures here? It seems unrelated as coming from the setup wizard...

lolautruche:
+1

yannickroger:
+1

PR #134, opened 2015-01-28
Remove legacy dependency

lolautruche:
> https://jira.ez.no/browse/EZP-23935

* [x] Activate legacy settings only if LegacyBundle is present
* [x] Use assets from DemoBundle / PlatformUIAssetsBundle
* [x] Remove website toolbar
* [x] Remove tag cloud
* [x] Only include legacy templates if legacy bridge is present
* [x] Add missing templates (without stuff relying on legacy)
* [x] Fix up legacy links
* [x] Implement generic line template


Resources/views/block_item/product.html.twig
```diff
@@ -0,0 +1,21 @@
...
+    </div>
+    <div class="product-info">
+        <div class="attribute-header">
+            <a href="{{ path( location ) }}" class="teaser-link">
+                <h3>{{ ez_content_name( content ) }}</h3>
+            </a>
+        </div>
+    </div>
+</article>
+
```
crevillo: so, we aren't showing price nor submit button here? i guess this is because shop still relies on legacy, right?

Resources/views/page_head_script.html.twig
```diff
@@ -1,12 +1,12 @@
...
 <script type="text/javascript">
-var YUI3_config = {base: "{{ asset( 'extension/ezjscore/design/standard/lib/yui/3.17.2/build/' ) }}", modules: {}, combine: false};
+var YUI3_config = {base: "{{ asset( 'bundles/ezplatformuiassets/vendors/yui3/build/' ) }}", modules: {}, combine: false};
 </script>
 
 {% javascripts
-    'extension/ezjscore/design/standard/lib/yui/3.17.2/build/yui/yui-min.js'
-    'extension/ezjscore/design/standard/javascript/jquery-1.10.2.min.js'
-    'extension/ezjscore/design/standard/javascript/jquery-ui-1.10.3.custom.min.js'
+    '@eZPlatformUIAssetsBundle/Resources/public/vendors/yui3/build/yui/yui-min.js'
```
crevillo: so, platform needs to be required now by demo too?

crevillo:
+0.5 :). i guess this is the way to go but at the end, demo will have less functionalities than it had before.

Resources/views/block_item/product.html.twig
lolautruche: yep

Resources/views/page_head_script.html.twig
lolautruche: platform assets bundle only

lolautruche:
Review ping @bdunogier @andrerom @dpobel @yannickroger

andrerom:
+0,1

however:
- we don't need jquery migration stuff here
- we can use jquery 2 here (but we can do that as followup maybe as part of boostrap 3 upgrade)
- ideally we should not put js assets in GIT*


* @dpobel Would it be ok if we rename asset bundle from PlatformUIAssets to PlatformAssets so it can contain scope for whole Platform? (also frontend) Or should we create a DemoUIAssetsBundle as well?

andrerom:
part 2:

Regarding what Carlos said, would make sense to keep the attribute view of show/other field types even if not part of Core.

Reason: As long as they are configured as nullType which you have to to avoid exception, ez_is_field_empty will return true so will avoid displaying field type.

lolautruche:
> Regarding what Carlos said, would make sense to keep the attribute view of show/other field types even if not part of Core.
> 
> Reason: As long as they are configured as nullType which you have to to avoid exception, ez_is_field_empty will return true so will avoid displaying field type.

This is not as easy... As for price, one solution would be to require PriceBundle, but for displaying forms it's a bit more complicated...

andrerom:
why do we need to require Pricebundle? by default it is NullType, so template code should be able to be there w/o issues. And then if you install Pricebundle price will show up. Shop functionality (basket) is a completely other topic of course.

bdunogier:
> Why would we need to require Pricebundle?

It only depends on the scope. If null mapping is acceptable, it's clearly easier.

andrerom:
Besides BDD failures: +1

@crevillo the changes I propose above can be done as separate PR.

crevillo:
as said imho, from a dev point of view what's done here makes totally sense. my worry is about ez installers, installing demobundle and not having all the things they had in previous versions. that's all :)

andrerom:
> my worry is about ez installers

we will deal with that (expectations), with legacy gone some features will go away for a while, this is the nature of things and with Platform it becomes more clear what is part of new stack and not yet.

crevillo:
ok, +1 :)

dpobel:
> * @dpobel Would it be ok if we rename asset bundle from PlatformUIAssets to PlatformAssets so it can contain scope for whole Platform? (also frontend) Or should we create a DemoUIAssetsBundle as well?

I would prefer to not mix the assets and keep PlatformUIAssetBundle for PlatformUI to keep it clean and focused on one thing. That said, I don't know if we really need a DemoUIAssetsBundle, if we are about to clean the DemoBundle to use Boostrap 3, DemoBundle will only need Bootstrap and jQuery which can be embed in it directly since it's quite small and does not need to be updated very often.

bdunogier:
> I would prefer to not mix the assets and keep PlatformUIAssetBundle for PlatformUI to keep it clean and focused on one thing.

Agreed.

andrerom:
> DemoBundle will only need Bootstrap and jQuery which can be embed in it directly since it's quite small and does not need to be updated very often.

true, but do use jquery UI atm?
As soon as we merge these things they are part of the size of the repo forever, so legacy repo (in mini scale) :)

andrerom:
After a short kitchen discussion, as we aim to try to stay up to date with boostrap and jquery going forward, could we create a DemoAssetBundle which pulls in jquery2 ? (boostrap 3 will be added alter)

This implies that we don't really need jQuery UI and jQuery migration stuff here, so alternative is that we just remove that in this PR and do the asset stuff later (but it is not hard to do, see [here](https://github.com/ezsystems/PlatformUIAssetsBundle))

lolautruche:
Tests are passing. Merging.

PR #118, opened 2014-11-03
EZP-23546: adding language switcher to DemoBundle

flovntp:
Language switcher was missing in the demoBundle.
I just added it into DemoBundle for eng <-> fre siteaccess as an example.
There is no design (for bootstrap3, please see http://share.ez.no/blogs/huck-florent/language-switcher-using-bootstrap3/(language)/eng-GB). 
It only display a list with flags and siteaccesses name on the top right corner.

Link: https://jira.ez.no/browse/EZP-23546

PR #135, opened 2015-02-02
EZP-23546: adding language switcher to DemoBundle

yannickroger:
Link: https://jira.ez.no/browse/EZP-23546

## Description
This is a cleaned up version of #118 by @flovntp. I left the credit of the commit to him just in case I introduced a nasty bug… hum… no no… to reward his great work and commitment :)

## Test
Manual test.


Controller/LanguageSwitcherController.php
```diff
@@ -0,0 +1,60 @@
...
+    {
+        /** @var \eZ\Publish\Core\Helper\TranslationHelper $translationHelper */
+        $translationHelper = $translationHelper = $this->container->get( 'ezpublish.translation_helper' );
+        $siteaccess = [];
+        $availableLanguages = [];
+
+        // create an array for corresponding siteaccesses names depending on the lang
+        foreach ( $translationHelper->getAvailableLanguages() as $lang )
+        {
+            if ( $lang != null )
```
crevillo: when can the lang be null?

Controller/LanguageSwitcherController.php
```diff
@@ -0,0 +1,60 @@
...
+                $availableLanguages[] = $lang;
+            }
+        }
+
+        // get current eZ language
+        $currentSFLanguage = $request->get( '_locale' );
+        $currentEzLanguage = array_search(
+            $currentSFLanguage,
+            $this->container->getParameter( 'ezpublish.locale.conversion_map' )
+        );
```
lolautruche: Why don't you just use `ezpublish.locale.converter` service?

andrerom:
Besides comment by @crevillo, +1

Resources/config/services.yml
```diff
@@ -6,6 +6,7 @@ parameters:
     ezdemo.email_helper.class: EzSystems\DemoBundle\Helper\EmailHelper
     ezdemo.form.type.feedback.class: EzSystems\DemoBundle\Form\Type\FeedbackType
     ezdemo.form.type.simple_search.class: EzSystems\DemoBundle\Form\Type\SimpleSearchType
+    ezdemo.controller.language_switcher.class: EzSystems\DemoBundle\Controller\LanguageSwitcherController
```
yannickroger: This is useless, I will remove it.

Resources/views/parts/languages_switcher.html.twig
```diff
@@ -0,0 +1,22 @@
...
+    {# Looping over all available languages to display the links #}
+
+    {% for lang in availableLanguages %}
+        {# This time, we alter the "siteaccess" parameter directly. #}
+        {# We get the right siteaccess with the help of ezpublish.translationSiteAccess() helper #}
+
+        {% do routeRef.set( "siteaccess", ezpublish.translationSiteAccess( lang ) ) %}
+        {% if lang != currentLanguage %}
+            <a href="{{ url( routeRef ) }}" title="{{ 'Switch to %siteaccess%'|trans({ '%siteaccess%': siteaccess[lang] })  }}">
+                <img src="{{ asset( '/share/icons/flags/' ~ lang ~ '.gif' ) }}" title="{{ lang }}">
```
lolautruche: Please don't rely on legacy assets. Better import them in the DemoBundle itself.

Controller/LanguageSwitcherController.php
yannickroger: Actually it happens a lot because there is a fallback of the default configuration returning an empty array. Maybe there is something we could fix/document (in case it's my configuration that is not right) ? ping @lolautruche

Controller/LanguageSwitcherController.php
```diff
@@ -0,0 +1,60 @@
...
+     * @param RouteReference $routeReference
+     *
+     * Displays the language switcher
+     *
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function showSwitcherAction( Request $request, RouteReference $routeReference )
+    {
+        /** @var \eZ\Publish\Core\Helper\TranslationHelper $translationHelper */
+        $translationHelper = $translationHelper = $this->container->get( 'ezpublish.translation_helper' );
```
yannickroger: fail

Controller/LanguageSwitcherController.php
lolautruche: I don't get it

Resources/views/parts/languages_switcher.html.twig
yannickroger: Should I add only fre and eng flags or all of them ?

Controller/LanguageSwitcherController.php
yannickroger: If you have only language installed SA (for example eng) and all the default setting after running the wizard, `$translationHelper->getAvailableLanguages()` will return:

```
array (size=2)
  0 => null
  1 => string 'eng-GB' (length=6)
```

crevillo: doh... weird, isn't it?

Resources/views/parts/languages_switcher.html.twig
andrerom: what about this one: https://packagist.org/packages/components/flag-icon-css

andrerom:
Regarding flags add eng, fre, ger and nor (enough for alpha1 demos) and we deal with the proper solution later.

crevillo:
for later, i like @andrerom proposal about using that external package for the flags.

yannickroger:
@crevillo it would be great to have the flags within the kernel, as demobundle is not meant to be included in projects. That's one of the reason we don't add it now.

andrerom:
we don't need to add it in the kernel if we can reuse some external package, anyway, get some flags in here to remove the legacy use :) Throw in spanish as well if you see this before you are done ;)

yannickroger:
is @crevillo buying beers if we add spanish ? ;)

andrerom:
I think we owe him beers already, but maybe that takes down the backlog a bit ;)

crevillo:
the idea using an external package is lazyness mainly :). Just saying... who knows if we could need a Catalonia flag soon :)
Yes, beers for everybody. i owe our croatian friends some of them too :) /cc @emodric

yannickroger:
Flags added. 
@andrerom  Should I create some kind of follow up issue for the `$translationHelper->getAvailableLanguages()`  problem mentionned in the comment above ?

andrerom:
@yannickroger yes please, it looks strange but should be handled separately.

yannickroger:
Issue created https://jira.ez.no/browse/EZP-23995

emodric:
> Yes, beers for everybody. i owe our croatian friends some of them too :) /cc @emodric

Beers all around, yes! \o/

PR #136, opened 2015-02-04
EZP-22361: Extracted and updated Bootstrap to newest version 3.3.2

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-22361

This PR requires #143.


bdunogier:
Why are there controller methods and helpers in a change about upgrading bootstrap ?

clash82:
They were needed to display [folder] content in Twig templates (which were only in legacy version).

michalpipa:
:+1:

Resources/public/less/buttons.less
```diff
@@ -126,7 +121,7 @@
 .defaultbutton,
 .defaultbutton:hover {
   text-shadow: 0 -1px 0 rgba(0,0,0,.25);
-  color: @white;
+  color: @white !important;
```
sunpietro: Important? That should be avoided.

Resources/public/less/comments.less
```diff
@@ -1,213 +1,179 @@
-/* COMMENT LIST */
-div.ezcom-view-list
-{
-    background-color: #ffffff;
+div.ezcom-view-list {
```
sunpietro: We should avoid using selectors like: `div.ezcom-view-list`.

Resources/public/less/comments.less
```diff
@@ -1,213 +1,179 @@
...
+  background-color: #ffffff;
 }
 
-div.ezcom-comment-title
-{
-    clear: both;
-    font-style: italic;
+div.ezcom-comment-title {
```
sunpietro: Same as https://github.com/ezsystems/DemoBundle/pull/136/files#r24088117

Resources/public/less/comments.less
```diff
@@ -1,213 +1,179 @@
...
 
-div.ezcom-field
-{
-    padding-left: 3px;
-    padding-right: 3px;
-    padding-top: 10px;
+div.ezcom-field {
+  padding-left: 3px;
+  padding-right: 3px;
+  padding-top: 10px;
```
sunpietro: Can't we use a shorthand for this?

Resources/public/less/comments.less
```diff
@@ -1,213 +1,179 @@
...
+}
 
-div.ezcom-page a
-{
-    padding: 1px;
-    background-color: #ffffff;
-    border: 1px solid #cccccc;
+div.ezcom-page a {
+  padding: 1px;
+  background-color: #ffffff;
```
sunpietro: Nitpick: you can use shorthand hex value for that :) like `#fff`.

Resources/public/less/comments.less
```diff
@@ -1,213 +1,179 @@
...
+/*---- PAGING ----*/
 /*---- ERROR MESSGE ----*/
-div.ezcom-delete .message-confirmation, div.ezcom-add-result .message-error
-{
-    margin-left: auto;
-    margin-right: auto;
+
+div.ezcom-delete .message-confirmation, div.ezcom-add-result .message-error {
+  margin-left: auto;
+  margin-right: auto;
```
sunpietro: Just `margin: auto`?

Resources/public/less/comments.less
```diff
@@ -1,213 +1,179 @@
...
 
-div.ezcom-setting-count
-{
-    float: right;
-    width: 20%;
-    display: inline
+div.ezcom-setting-select {
+  float: left;
+  width: 130px;
+  display: inline;
```
sunpietro: `display: inline`should be removed. Floated elements are always block elements.

Resources/public/less/comments.less
```diff
@@ -1,213 +1,179 @@
...
 }
 
-div.ezcom-setting-content
-{
-    display: inline;
+div.ezcom-setting-count {
+  float: right;
+  width: 20%;
+  display: inline
```
sunpietro: Same as https://github.com/ezsystems/DemoBundle/pull/136/files#r24088538

Resources/public/less/forms.less
```diff
@@ -1,587 +1,141 @@
...
-  margin-bottom: 0;
-  vertical-align: middle;
-}
-.radio.inline + .radio.inline,
-.checkbox.inline + .checkbox.inline {
-  margin-left: 10px; // space out consecutive inline controls
+  border-radius: 4px;
+  padding: 4px;
+  width: 220px;
+  _display: block;
```
sunpietro: Is that really needed?

Resources/public/less/forms.less
```diff
@@ -1,587 +1,141 @@
...
-    *margin-top: 1px; /* IE6-7 */
+  label {
+    font-weight: bold;
   }
-}
-.input-append {
-  input,
-  .uneditable-input {
+  .element,
+  .halfelement {
```
sunpietro: What is `.halfelement`?

bdunogier:
> They were needed to display [folder] content in Twig templates (which were only in legacy version).

Well, then they should be in another pull request. It's not acceptable that controller actions and configuration are added as a side effect of a pull-request without even being mentioned in the JIRA Issue and or/pull request.

Overall, keeping matters separated simplifies things, and makes reviews more efficient.

I'd really prefer if you could move those to another branch, and have it discussed and merged separately.

Resources/public/less/forms.less
```diff
@@ -1,587 +1,141 @@
...
 
-.box {
-    width: 99%;
-}
-.halfbox {
-    width: 49%;
+input:focus:required:invalid, textarea:focus:required:invalid, select:focus:required:invalid {
+  color: #b94a48;
+  border-color: #ee5f5b;
+  &:focus {
```
sunpietro: Why?
The final will be `input:focus:required:invalid:focus`

Resources/public/less/full-gallery.less
```diff
@@ -0,0 +1,127 @@
...
+    z-index:10;
+    overflow:hidden;
+    padding-top:5px;
+    .cursor {
+      position:absolute;
+      top:0;
+      z-index:11;
+      width:33px;
+      display:inline-block;
+      display:none;
```
sunpietro: So `display: inline-block;` or `display: none;`?

Resources/public/less/full-gallery.less
```diff
@@ -0,0 +1,127 @@
...
+    .navig {
+      display:inline-block;
+      width:20px;
+      height:37px;
+      position:absolute;
+      top:31px;
+      z-index:11;
+    }
+    .next {
+      right:0px;
```
sunpietro: Nitpick: `px` can be omitted in such cases.

Resources/public/less/full-gallery.less
```diff
@@ -0,0 +1,127 @@
...
+      position:relative;
+      top:0;
+      left:0;
+      white-space:nowrap;
+      z-index:1;
+      font-size:10px;
+      li {
+        display:inline-block;
+        &:before {
+          display:none;
```
sunpietro: What for?

Resources/public/less/layouts.less
```diff
@@ -1,338 +1,339 @@
...
+    .login-form {
+      .box-shadow(0 5px 5px rgba(0,0,0,0.35));
+      background: @blue;
+      #gradient > .vertical(@blue, @blueDark);
+    }
+  }
+  .login-form {
+    width: 280px;
+    margin: 0;
+    color: white;
```
sunpietro: Color names should consistent. There should be a hex value: `#fff`

Resources/public/less/layouts.less
```diff
@@ -1,338 +1,339 @@
...
-        .login-form {
-            display: block;
-            height: 140px;
-            overflow: hidden;
-            bottom: -140px;
-        }
+      display: block;
+      height: 130px;
+      overflow: hidden;
+      bottom: -130px;
```
sunpietro: Nitpick: it can be written like that: `transform: translateY(100%);`. You won't have to care about the element's height value.

Resources/public/less/layouts.less
```diff
@@ -1,338 +1,339 @@
...
+  }
+  > .content-view-aside {
+    position: relative;
+    &:first-child {
+      background: -moz-linear-gradient(top,  rgba(255,255,255,1) 0px, rgba(255,255,255,0) 30px); /* FF3.6+ */
+      background: -webkit-gradient(linear, left top, left bottom, color-stop(0px,rgba(255,255,255,1)), color-stop(30px,rgba(255,255,255,0))); /* Chrome,Safari4+ */
+      background: -webkit-linear-gradient(top,  rgba(255,255,255,1) 0px,rgba(255,255,255,0) 30px); /* Chrome10+,Safari5.1+ */
+      background: -o-linear-gradient(top,  rgba(255,255,255,1) 0px,rgba(255,255,255,0) 30px); /* Opera 11.10+ */
+      background: -ms-linear-gradient(top,  rgba(255,255,255,1) 0px,rgba(255,255,255,0) 30px); /* IE10+ */
+      background: linear-gradient(top,  rgba(255,255,255,1) 0px,rgba(255,255,255,0) 30px); /* W3C */
```
sunpietro: What browsers should DemoBundle support?

Resources/public/less/layouts.less
```diff
@@ -1,338 +1,339 @@
...
+          margin-top: 5px;
         }
-        article,
-        .block-content {
+      }
+      &.selected-products {
+        padding: 20px 0;
+        .clearfix();
+        .attribute-image {
+          display: block;
```
sunpietro: `display: block` is not needed there.

Resources/public/less/layouts.less
```diff
@@ -1,338 +1,339 @@
...
+          float: left;
+          width: 25%;
+          img {
+            margin: 0 20px;
+          }
+        }
+        .product-info {
+          display: block;
```
sunpietro: Same here. See https://github.com/ezsystems/DemoBundle/pull/136/files#r24089605

Resources/public/less/layouts.less
```diff
@@ -526,527 +536,527 @@ aside {
...
+              height: 19px;
+              width: 19px;
+              margin-right: 0.5em;
+              display: inline-block;
+              #gradient > .vertical(@gray, @grayLow);
+              .border-radius(3px);
+              vertical-align: text-bottom;
+              position: relative;
+              left: auto;
+              top: auto;
```
sunpietro: Was that element absolutely positioned earlier? In fot, please remove `top`, `left` properties.

Resources/public/less/layouts.less
```diff
@@ -526,527 +536,527 @@ aside {
...
+  }
+  .col-md-2 {
+    text-align: right;
+  }
+  .form {
+    background-color: #f4f4f4;
+    padding: 10px 0 10px 0;
+    -webkit-border-radius: 10px;
+    -moz-border-radius: 10px;
+    border-radius: 10px;
```
sunpietro: Nitpick: you could use `.border-radius()` mixin

Resources/public/less/layouts.less
```diff
@@ -526,527 +536,527 @@ aside {
...
+      margin: 0;
+      padding: 0;
+      text-align: center;
+    }
+    .previous,
+    .next {
+      font-weight: bold;
+      margin: 0 4px 0 4px;
+      padding-top: 0.1em;
+      display: inline; /* Fix for IE */
```
sunpietro: What IE?

Resources/public/less/navbar.less
```diff
@@ -1,422 +1,201 @@
...
+      a {
+        display: inline-block;
+        padding: 5px 10px 5px 30px;
+      }
+      ul {
+        margin:0;
+        background-color: @blue;
+        position: absolute;
+        left: 170px;
+        top: -1000px;
```
sunpietro: ?

Resources/public/less/navbar.less
```diff
@@ -1,422 +1,201 @@
...
+  .navbar-header {
+    display: block;
   }
-}
+  .navbar-menu {
+    .container {
+      margin: 0;
+    }
+    ul {
+      float: none !important;
```
sunpietro: Important?

Resources/public/less/star-rating.less
```diff
@@ -1,21 +1,20 @@
...
+
 .ezsr-star-rating,
 .ezsr-star-rating a:hover,
 .ezsr-star-rating a:active,
 .ezsr-star-rating a:focus,
-.ezsr-star-rating .ezsr-current-rating
-{
-    background: url(../bundles/ezdemo/images/star.png) left -1000px repeat-x;
+.ezsr-star-rating .ezsr-current-rating {
+  background: url(../bundles/ezdemo/images/star.png) left -1000px repeat-x;
```
sunpietro: ?

Resources/public/less/star-rating.less
```diff
@@ -25,74 +24,70 @@
...
   background-position: left top;
 }
-.ezsr-star-rating li
-{
+
+.ezsr-star-rating li {
   display: inline;
   margin: 0;
   padding: 0;
   position: static;
```
sunpietro: `position: static` is a default browser value for any HTML element

Resources/public/less/tables.less
```diff
@@ -124,51 +85,21 @@ table.list {
       background-color: #f9f9f9;
     }
     td, th {
-        border-color: transparent;
+      border-color: transparent;
```
sunpietro: If border is transparent, then do we realy need any border? Maybe: `border: 0 none;` would be a better option?

Resources/public/less/type.less
```diff
@@ -289,20 +168,20 @@ blockquote {
...
-                content: '\BB';
-            }
-        }
+    .clearfix();
+    p {
+      margin-bottom: 20px;
+      font-family: inherit;
+      font-size: inherit;
+      line-height: inherit;
+      color: inherit;
```
sunpietro: ?

Resources/public/less/utilities.less
```diff
@@ -22,68 +22,64 @@
...
 
-// Visibility
-.invisible {
-  visibility: hidden;
+.hidden {
+  position: absolute;
+  overflow: hidden;
+  clip: rect(0 0 0 0);
+  height: 1px; width: 1px;
+  margin: -1px; padding: 0; border: 0;
```
sunpietro: Each property should be in separate line.

Resources/public/less/website-toolbar.less
```diff
@@ -1,6 +1,11 @@
...
-    vertical-align: baseline;
+  vertical-align: baseline;
 }
+
 #ezwt-content input[disabled] {
-    background-color: transparent;
+  background-color: transparent;
 }
+
+#ezwt a img {
```
sunpietro: Can we avoid ID's?

Resources/views/full/folder.html.twig
```diff
@@ -0,0 +1,50 @@
...
+                                    {% endfor %}
+                                </ul>
+                            {% endif %}
+                        </section>
+                    </aside>
+                </div>
+            </div>
+        </div>
+    </section>
+{% endblock %}
```
sunpietro: It's a commentabout the whole file. I don't like the HTML code structure. Where can I see the final result?

Resources/views/line/folder.html.twig
```diff
@@ -0,0 +1,7 @@
+{% if subLinks|length > 0 %}
+    <ul>
```
sunpietro: You could create some folder-view-specific CSS classes to make CSS styling easier.

Resources/views/page_header_logo.html.twig
```diff
@@ -1,3 +1,3 @@
-<div class="span8">
-    <a href="{{ path( ezpublish.rootLocation ) }}" title="" class="logo"><img src="{{ asset('bundles/ezdemo/images/logo-ez.png') }}" alt="" /></a>
+<div class="col-md-8 col-xs-6 col-sm-6">
```
sunpietro: `col-xs-6 col-md-8` is more than enough. `col-xs-6` is considered first when rendering and when requirements for `col-md-8` are met then new CSS styles applies.

sunpietro: `col-xs-6 col-md-8` is more than enough. `col-xs-6` is considered first when rendering and when requirements for `col-md-8` are met then new CSS styles applies.

Resources/views/page_header_searchbox.html.twig
```diff
@@ -1,7 +1,7 @@
-{{ form_start( form, { 'action': path('ezpublish_ezdemo_search'), 'attr': { 'class': 'span4 form-search' } } ) }}
+{{ form_start( form, { 'action': path('ezpublish_ezdemo_search'), 'attr': { 'class': 'col-md-4 col-xs-6 col-sm-6 form-search' } } ) }}
```
sunpietro: Same as https://github.com/ezsystems/DemoBundle/pull/136/files#r24092275

Resources/views/page_topmenu.html.twig
```diff
@@ -1,23 +1,21 @@
...
 
-{% if submenu is defined %}
-<div class="navbar-inner sublevel">
-    <div class="container">
-        <div class="nav-collapse">
-            {{ knp_menu_render(submenu, {'firstClass': 'firstli', 'lastClass': 'lastli', 'branch_class': 'nav'}) }}
+    {% if submenu is defined %}
+        <div class="navbar-collapse collapse navbar-menu navbar-sublevel">
+            <div class="container">
+                {{ knp_menu_render(submenu, {'firstClass': 'firstli', 'lastClass': 'lastli', 'branch_class': 'nav'}) }}
```
sunpietro: I don't like CSS class names. The could be styled with CSS pseudoselectors like `:first-child`

bdunogier:
In case you haven't noticed, I had posted something above Piotr's 26 comments ;-)

I'll paste it here:


> (about the controller methods) They were needed to display [folder] content in Twig templates (which were only in legacy version).

Well, then they should be in another pull request. It's not acceptable that controller actions and configuration are added as a side effect of a pull-request without even being mentioned in the JIRA Issue and or/pull request.

Overall, keeping matters separated simplifies things, and makes reviews more efficient. I'd really prefer if you could move those to another branch, and have it discussed and merged separately.

The situation above, where comments get burried below many other, is also a consequence of an overscoped pull-request.

bdunogier:
Oh, one more thing. I'm clearly not an expert in this area, but I see quite a few new CSS classes. Are they fixes required by the new bootstrap version ? Or do they have visual consequences ?

clash82:
@Piotr: OMG, thanks for a good work! ;-)

@Bertrand: Don't worry, I've seen your comment :-) Today I'm gonna split this PR into separate parts...
Which CSS classes exactly you mean? Bootstrap v3 has a lots of new features which are requiring new classes.

dpobel:
It would be nice to not embed Bootstrap in the DemoBundle. We discussed a bit that topic already in https://github.com/ezsystems/DemoBundle/pull/134 and the outcome was to create a DemoAssetsBundle so such things. But I've just realized that we can already add Bootstrap as a dependency in composer.json without creating an eZ Bundle for that (either https://packagist.org/packages/twitter/bootstrap or https://packagist.org/packages/components/bootstrap depending on what we want/need).
@lolautruche @andrerom @bdunogier any opinion on that ?

lserwatka:
I agree with @dpobel here, Bootstrap should be externalized in this case, it is even better if there are already packages for it.

Resources/public/less/navbar.less
clash82: Hmm...
DemoBundle doesn't have language selector enabled - is this a deprecated or hidden feature?!

andrerom:
re externalizing

lets also do this for jquery then, resolving the topic we had in #134 on that.
ref: https://packagist.org/packages/components/jquery

andrerom:
@clash82 & @lserwatka  !! Please make sure to rebase history on this so there is no bootstrap what so ever in history!!

Otherwise it will *forever* be part of size of the repository (relevant when you checkout), as happened #134 with jQuery UI.

bdunogier:
> Please make sure to rebase history on this PR so there is no bootstrap what so ever in history!!
+5

To be frank, it might be easier to cherry pick the commits to new branches, and create new pull requests altogether.

clash82:
@andrerom should I commit compiled bootstrap.css file to repository?
This file is generated when compiling original Bootstrap and eZDemoBundle less files (custom styles).

andrerom:
@clash82 Good question, @dpobel? I guess we can't avoid that, and it's demo bundle specific so makes sense to have it here.

dpobel:
Well, it would nice to not commit it and to use the Lessphp Assetic filter to dynamically build the resulting CSS file. I just hope Lessphp is advanced enough to compile our less files.

clash82:
@andrerom are you agree with @dpobel ?

andrerom:
@clash82 I agree that we should ideally aim for that, but if it does not work we should rather just put that here for now ( I'm assuming the files won't be that big, please provide size info if we go this route ;) )

clash82:
New version with rebased commits. BootstrapJS and jQuery (as a dependency) was moved to composer. I'm planning to try add some code to composer so it will be able to compile less files after installing.

andrerom:
Cool, as for the less compilation, which command is that? assetic:dump ?
If so we already use it and only need to configure it for ezpublish-community

Resources/views/line/folder.html.twig
andrerom: If this is used instead of `Resources/views/line/line.html.twig` then you should follow the same setup, but this seems to be something else or?

Resources/views/line/folder.html.twig
```diff
@@ -0,0 +1,7 @@
...
+        {% for link in subLinks %}
+            <li><a href="{{ path( 'ez_urlalias', { 'locationId': link.contentInfo.mainLocationId } ) }}" title="{{ ez_field_value( link, 'name' ) }}">{{ ez_field_value( link, 'name' ) }}</a></li>
+        {% endfor %}
+    </ul>
+{% endif %}
```
andrerom: Always new line at end of code files, change your IDE to enable this ;)

Resources/views/page_header_links.html.twig
```diff
@@ -16,7 +16,7 @@
     <li id="login" class="transition-showed">
         <a href="#login" title="show login form" class="show-login-form">{{ 'Login'|trans }}</a>
         <a href="#" title="hide login form" class="hide-login-form">{{ 'Login'|trans }}</a>
-        <form class="login-form span3" action="{{ path( 'login_check' ) }}" method="post">
+        <form class="login-form" action="{{ path( 'login_check' ) }}" method="post">
```
andrerom: No need for the size class anymore?

composer.json
```diff
@@ -15,7 +15,8 @@
         "ezsystems/comments-bundle": "@dev",
         "ezsystems/platform-ui-assets-bundle": "~0.1",
         "whiteoctober/breadcrumbs-bundle": "~1.0.1",
-        "knplabs/knp-menu-bundle": "1.*"
+        "knplabs/knp-menu-bundle": "1.*",
+        "components/bootstrap": "3.3.2"
```
andrerom: Should be set to "~3.3.2" so we get patch updates without having to change anything.

Resources/public/less/website-toolbar.less
andrerom: @clash82 / @lserwatka I think we removed ezwt recently, @lolautruche might be able to confirm otherwise check the legacy removal PR that was merged. (ezwt is purely legacy)

clash82:
@andrerom Firstly we need to use LessPHP (http://leafo.net/lessphp/) to compile Less to CSS on-the-fly, then assetic:dump can go to action :-)

composer.json
clash82: It may be a bit risky because some of the DemoBundle CSS classes strongly interacts with BootstrapJS core classes. I would prefer to update it manually (?)

Resources/views/line/folder.html.twig
clash82: Sorry :-)

composer.json
andrerom: Ok, up to you :)
I didn't see anything risky in [3.3.0 to 3.3.2](https://github.com/twbs/bootstrap/compare/v3.3.0...v3.3.2#diff-9476f0f857cc69f29e7cdb7865d409e2) update, and this being a demo bundle I thought we could threat it like we do with php packages, but lets avoid breaks.

andrerom:
> @andrerom Firstly we need to use LessPHP (http://leafo.net/lessphp/) to compile Less to CSS on-the-fly, then assetic:dump can go to action :-)

Ok, something your confident we can have in place before Tuesday? Have you tested it yet?

andrerom:
As agreed with @lserwatka, lets skip lessPHP if it is not very well maintained, lets keep on generating assets in this PR as we already where doing that anyway, and rather consider to do followup having a asset bundle line Platform UI later. 

So besides the line folder template which should probably be in the other PR under a different name, what else is remaining here?

clash82:
@andrerom everything should be ready for now.

andrerom:
Tests don't execute as assetic fails with:
```                                                                           
  [RuntimeException]                                                           
  The source file "/home/travis/build/ezpublish-community/ezpublish/../vendor  
  /components/jquery/jquery.min.js" does not exist.     
```

andrerom:
@clash82 looked into issue above?

clash82:
@andrerom I'm working as fast as I can :-)

andrerom:
:)

clash82:
@andrerom Unfortunately there isn't a quick solution for assetic fails problem. There are different vendor paths according to working environment (test, prod). This situation must be discussed.

andrerom:
ok, but your other PR can be made ready for final review right? I anticipate some merge conflicts with all the 4 incoming PR's, so need to start soon :)

clash82:
@andrerom i think it's ready for final review :-)

andrerom:
not really, the css and less changes are still in the PR.

clash82:
@andrerom I think they should stay because they are part of this task, but maybe you have something else on your mind?

andrerom:
I was referring to #137, but that is fixed now.

</end-generic-pr-ordering-discussion>

clash82:
Is there a small chance to merge this PR? :-)

andrerom:
Well, tests should be passing, so please rebase on top of latests master so we can see how the tests responds to that.

clash82:
Still not passing:

`[Behat\Behat\Context\Exception\ContextNotFoundException]
'eZ\Bundle\EzPublishLegacyBundle\Features\Context\SetupWizard\Context' context class not found and can not be used.`

andrerom:
My bad, I should have realized it was broken after the installer merge like BehatBundle was. You can retry now, or when/if 5a0133a5167fcb26ec019 passes: https://travis-ci.org/ezsystems/DemoBundle/builds/51440583

clash82:
That's weird... responsive.css was removed from `page-head_style.html.twig`...

andrerom:
You can try adding this to the PR:
```diff
diff --git a/.travis.yml b/.travis.yml
index f8c37f5..c49e7ab 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,7 @@ before_script:
 
   # Replace demobundle with the one from pull-request
   - rm -rf vendor/ezsystems/demobundle/EzSystems/DemoBundle && mv "$BRANCH_BUILD_DIR" vendor/ezsystems/demobundle/EzSystems/DemoBundle
+  - php ezpublish/console cache:clear --env=behat --no-debug
   - php ezpublish/console assetic:dump --env=behat --no-debug
 
 # execute behat as the script command
```

clash82:
`[RuntimeException]
The source file "/home/travis/build/ezpublish-community/ezpublish/../vendor
/components/jquery/jquery.min.js" does not exist.`

OK, I know what's wrong.

I've added BootstrapJS dependency to DemoBundle's `composer.json`. By default (in test environment) composer installs only dependency from dev-master's `composer.json`, so the new version (in this PR)  is not included... and this is why BootstrapJS and JQuery are missing.

Best way to bypass this problem would be to update DemoBundle at the end of creation of test environment (after replacing dev-master's DemoBundle). Shortly: install DemoBundle twice in the same task.

What do you think?
ping @andrerom @lserwatka @michalpipa @dpobel @bdunogier

crevillo:
install a thing twice in the same task? doesn't sound very good... :/

clash82:
@crevillo it was only suggestion ;-)

Second solution: move BootstrapJS dependency to new PR and merge it.

michalpipa:
> Second solution: move BootstrapJS dependency to new PR and merge it.

:+1:

clash82:
ping @andrerom

andrerom:
ok, yes it is problem with how we inject current branch a bit brute force at the moment..

bdunogier:
@clash82 unless I misunderstand your problem, we've been facing this "problem" (more of an annoyance) for a while now.

What I do (others do as well is create a custom ezpublish-community branch (that won't be merged).

This branch changes the requirement of `demobundle:dev-master` to `demobundle:dev-YourDemoBundleBranch` in `composer.json`.

In the Demobundle PR, `.travis.yml` should be changed to checkout your custom branch instead of master. This commit must be removed before merging.

But we should really try to make this easier and smoother in the near future.

clash82:
@dpobel thanks for reply. Your solution seems to be a bit complicated, is it really necessary?

andrerom:
#143 has been merged, for next time either pick @bdunogier's ( not @dpobel ;) ) suggested way to run the test, or we can try to fix the issue with some travis effort.

.travis.yml
```diff
@@ -38,6 +38,7 @@ before_script:
 
   # Replace demobundle with the one from pull-request
   - rm -rf vendor/ezsystems/demobundle/EzSystems/DemoBundle && mv "$BRANCH_BUILD_DIR" vendor/ezsystems/demobundle/EzSystems/DemoBundle
+  - php ezpublish/console cache:clear --env=behat --no-debug
```
andrerom: can be removed now that boostrap dep. has been merged.

clash82: Cache must be cleared because it includes reference to `responsive.css`

clash82:
Yes we can!
@andrerom my mistake, sorry for that. They are both on the list when I press "b" ;-)

.travis.yml
andrerom: ok

andrerom:
+1

review ping @bdunogier @yannickroger @dpobel @lserwatka

PR #137, opened 2015-02-05
Missing PHP code for [folder] content

clash82:
This PR includes missing PHP code for displaying Twig version of [folder] content. See also #136

### DIFF
--- a/Controller/FolderController.php
+++ b/Controller/FolderController.php
@@ -1,0 +1,156 @@
+<?php
+/**
+ * This file is part of the ezsystems/demo-bundle package
+ *
+ * @copyright Copyright (C) eZ Systems AS. All rights reserved.
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ * @version //autogentag//
+ */
+
+namespace EzSystems\DemoBundle\Controller;
+
+use eZ\Publish\API\Repository\Values\Content\LocationQuery;
+use eZ\Publish\API\Repository\Values\Content\Query;
+use eZ\Publish\API\Repository\Values\Content\Query\SortClause;
+use eZ\Publish\API\Repository\Values\Content\Location;
+use eZ\Publish\Core\Pagination\Pagerfanta\ContentSearchAdapter;
+use eZ\Bundle\EzPublishCoreBundle\Controller;
+use Pagerfanta\Pagerfanta;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\HttpFoundation\Request;
+
+class FolderController extends Controller
+{
+    /**
+     * Displays the sub folder if it exists
+     *
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location of a folder
+     * @throws NotFoundHttpException $location is flagged as invisible
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function showFolderListAsideViewAction( Location $location )
+    {
+        $response = new Response();
+
+        if ( $location->invisible )
+        {
+            throw new NotFoundHttpException( "Location #$location->id cannot be displayed as it is flagged as invisible." );
+        }
+
+        $response->setSharedMaxAge( $this->getConfigResolver()->getParameter( 'content.default_ttl' ) );
+        $response->headers->set( 'X-Location-Id', $location->id );
+        $response->setVary( 'X-User-Hash' );
+
+        $languages = $this->getConfigResolver()->getParameter( 'languages' );
+
+        $includedContentTypeIdentifiers = $this->container->getParameter( 'ezdemo.folder.folder_tree.included_content_types' );
+
+        $subContentCriteria = $this->get( 'ezdemo.criteria_helper' )->generateSubContentCriterion(
+            $location, $includedContentTypeIdentifiers, $languages
+        );
+
+        $subContentQuery = new LocationQuery();
+        $subContentQuery->criterion = $subContentCriteria;
+        $subContentQuery->sortClauses = array(
+            new SortClause\ContentName()
+        );
+
+        $searchService = $this->getRepository()->getSearchService();
+        $subContent = $searchService->findLocations( $subContentQuery );
+
+        $treeChildItems = array();
+        foreach ( $subContent->searchHits as $hit )
+        {
+            $treeChildItems[] = $hit->valueObject;
+        }
+
+        return $this->get( 'ez_content' )->viewLocation(
+            $location->id,
+            'aside_sub',
+            true,
+            ['treeChildItems' => $treeChildItems]
+        );
+    }
+
+    /**
+     * Displays the list of article
+     *
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location of a folder
+     * @param \Symfony\Component\HttpFoundation\Request $request request object
+     * @throws NotFoundHttpException $location is flagged as invisible
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function showFolderListAction( Request $request, Location $location )
+    {
+        $response = new Response();
+
+        if ( $location->invisible )
+        {
+            throw new NotFoundHttpException( "Location #$location->id cannot be displayed as it is flagged as invisible." );
+        }
+
+        $response->setSharedMaxAge( $this->getConfigResolver()->getParameter( 'content.default_ttl' ) );
+        $response->headers->set( 'X-Location-Id', $location->id );
+        $response->setVary( 'X-User-Hash' );
+
+        $content = $this->getRepository()
+            ->getContentService()
+            ->loadContentByContentInfo( $location->getContentInfo() );
+
+        // Getting language for the current siteaccess
+        $languages = $this->getConfigResolver()->getParameter( 'languages' );
+
+        $excludedContentTypes = $this->container->getParameter( 'ezdemo.folder.folder_view.excluded_content_types' );
+
+        // Using the criteria helper (a demobundle custom service) to generate our query's criteria.
+        // This is a good practice in order to have less code in your controller.
+        $criteria = $this->get( 'ezdemo.criteria_helper' )->generateListFolderCriterion(
+            $location, $excludedContentTypes, $languages
+        );
+
+        // Generating query
+        $query = new LocationQuery();
+        $query->criterion = $criteria;
+        $query->sortClauses = array(
+            new SortClause\DatePublished()
+        );
+
+        // Initialize pagination.
+        $pager = new Pagerfanta(
+            new ContentSearchAdapter( $query, $this->getRepository()->getSearchService() )
+        );
+
+        $pager->setMaxPerPage( $this->container->getParameter( 'ezdemo.folder.folder_list.limit' ) );
+        $pager->setCurrentPage( $request->get( 'page', 1 ) );
+
+        $includedContentTypeIdentifiers = $this->container->getParameter( 'ezdemo.folder.folder_tree.included_content_types' );
+
+        // Get sub folder structure
+        $subContentCriteria = $this->get( 'ezdemo.criteria_helper' )->generateSubContentCriterion(
+            $location, $includedContentTypeIdentifiers, $languages
+        );
+
+        $subContentQuery = new LocationQuery();
+        $subContentQuery->criterion = $subContentCriteria;
+        $subContentQuery->sortClauses = array(
+            new SortClause\ContentName()
+        );
+
+        $searchService = $this->getRepository()->getSearchService();
+        $subContent = $searchService->findLocations( $subContentQuery );
+
+        $treeItems = array();
+        foreach ( $subContent->searchHits as $hit )
+        {
+            $treeItems[] = $hit->valueObject;
+        }
+
+        return $this->get( 'ez_content' )->viewLocation(
+            $location->id,
+            'full',
+            true,
+            ['pagerFolder' => $pager, 'treeItems' => $treeItems]
+        );
+    }
+}
--- a/Helper/CriteriaHelper.php
+++ b/Helper/CriteriaHelper.php
@@ -21,6 +21,46 @@ use DateInterval;
 class CriteriaHelper
 {
     /**
+     * Generate criterion list to be used to list article
+     *
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location Location of the folder
+     * @param string[] $excludeContentTypeIdentifiers Array of excluded contentType identifiers
+     * @param string[] $languages Array of languages
+     *
+     * @return \eZ\Publish\API\Repository\Values\Content\Query\Criterion
+     */
+    public function generateListFolderCriterion( Location $location, array $excludeContentTypeIdentifiers = array(), array $languages = array() )
+    {
+        $criteria = array();
+        $criteria[] = new Criterion\Visibility( Criterion\Visibility::VISIBLE );
+        $criteria[] = new Criterion\ParentLocationId( $location->id );
+        $criteria[] = new Criterion\LogicalNot( new Criterion\ContentTypeIdentifier( $excludeContentTypeIdentifiers ) );
+        $criteria[] = new Criterion\LanguageCode( $languages );
+
+        return new Criterion\LogicalAnd( $criteria );
+    }
+
+    /**
+     * Generate criterion list to be used to list sub folder items
+     *
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location Location of the folder
+     * @param string[] $includedContentTypeIdentifiers Array of included contentType identifiers
+     * @param string[] $languages Array of languages
+     *
+     * @return \eZ\Publish\API\Repository\Values\Content\Query\Criterion
+     */
+    public function generateSubContentCriterion( Location $location, array $includedContentTypeIdentifiers = array(), array $languages = array() )
+    {
+        $criteria = array();
+        $criteria[] = new Criterion\Visibility( Criterion\Visibility::VISIBLE );
+        $criteria[] = new Criterion\ContentTypeIdentifier( $includedContentTypeIdentifiers );
+        $criteria[] = new Criterion\LanguageCode( $languages );
+        $criteria[] = new Criterion\ParentLocationId( $location->id );
+
+        return new Criterion\LogicalAnd( $criteria );
+    }
+
+    /**
      * Generates an exclude criterion based on contentType identifiers.
      *
      * @param array $excludeContentTypeIdentifiers
--- a/Resources/config/default_settings.yml
+++ b/Resources/config/default_settings.yml
@@ -8,6 +8,12 @@ parameters:
     ezdemo.footer.latest_content.content_types: [article]
     # Number of blog_posts displayed per page in the blog view
     ezdemo.blog.blog_post_list.limit: 10
+    # Number of article displayed per page in the folder view
+    ezdemo.folder.folder_list.limit: 10
+    # ContentType identifiers we want to exclude from content list in folder view
+    ezdemo.folder.folder_view.excluded_content_types: [folder, image]
+    # ContentType identifiers we want to have in folder's tree
+    ezdemo.folder.folder_tree.included_content_types: [folder]
     # Decide whether to show or hide article summary in the full view
     ezdemo.article.full_view.show_summary: true
     # Decide whether to show or hide article image in the full view
--- a/Resources/config/ezdemo.yml
+++ b/Resources/config/ezdemo.yml
@@ -15,6 +15,19 @@ system:
             layout: eZDemoBundle::pagelayout.html.twig
             login_template: eZDemoBundle:Security:login.html.twig
         location_view:
+            aside:
+                folder:
+                    template: "eZDemoBundle:aside:folder.html.twig"
+                    match:
+                        Identifier\ContentType: [folder]
+
+            aside_sub:
+                folder:
+                    controller: "eZDemoBundle:Folder:showFolderListAsideView"
+                    template: "eZDemoBundle:aside_sub:folder.html.twig"
+                    match:
+                        Identifier\ContentType: [folder]
+
             full:
                 article:
                     controller: "eZDemoBundle:Demo:showArticle"
@@ -59,6 +72,11 @@ system:
                     template: "eZDemoBundle:full:call_to_action.html.twig"
                     match:
                         Identifier\ContentType: "call_to_action"
+                folder:
+                    controller: "eZDemoBundle:Folder:showFolderList"
+                    template: "eZDemoBundle:full:folder.html.twig"
+                    match:
+                        Identifier\ContentType: [folder]
 
             line:
                 article:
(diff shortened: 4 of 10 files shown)


yannickroger:
I think we could put this in a dedicated controller (at some point we'll have to refactor the demo controller, it does too much stuff).
Also, as your code is mainly based on the blog feature, you could add some BDD testing (we just added them to Blog). You can ping me or @bdunogier for a kickstart on that.

Except that, it looks good.

yannickroger:
As a side note I am not a big fan of the criteria helper and I hope we can refactor that as well at some point, but it's not the scope of this PR ;)

yannickroger:
Also rename your commit as it not only introduces a twig template but also controllers. "Implemented display of folders" something like that

Controller/DemoController.php
```diff
@@ -21,6 +21,126 @@
...
+        );
+
+        $searchService = $this->getRepository()->getSearchService();
+        $subContent = $searchService->findContent( $subContentQuery );
+
+        foreach ($subContent->searchHits as $hit)
+            $subLinks[] = $hit->valueObject;
+
+        return $this->render(
+            'eZDemoBundle:line:folder.html.twig',
```
andrerom: is it on purpose that this template is in the other PR?

Controller/DemoController.php
```diff
@@ -21,6 +21,126 @@
...
+        $subContentQuery->criterion = $subContentCriteria;
+        $subContentQuery->sortClauses = array(
+            new SortClause\ContentName()
+        );
+
+        $searchService = $this->getRepository()->getSearchService();
+        $subContent = $searchService->findContent( $subContentQuery );
+
+        $subContentObjects = array();
+        foreach ($subContent->searchHits as $hit)
```
crevillo: cs  after and before opening and closing parents

Controller/DemoController.php
```diff
@@ -21,6 +21,126 @@
...
+            $subContentObjects[] = $hit->valueObject;
```
crevillo: not totally necessary, but no brackets here? I think we prefer add brackets...

Controller/DemoController.php
```diff
@@ -21,6 +21,126 @@
...
+        $subContentQuery->criterion = $subContentCriteria;
+        $subContentQuery->sortClauses = array(
+            new SortClause\ContentName()
+        );
+
+        $searchService = $this->getRepository()->getSearchService();
+        $subContent = $searchService->findContent( $subContentQuery );
+
+        foreach ($subContent->searchHits as $hit)
+            $subLinks[] = $hit->valueObject;
```
crevillo: same as i said below about cs and brackets.

lserwatka:
We are expecting new update by @clash82 to this PR soon

Resources/views/full/folder.html.twig
```diff
@@ -0,0 +1,48 @@
...
+                <aside class="col-md-4">
+                    <section class="subnavigation row nav-collapse">
+                        <div class="attribute-header">
+                            <h2>{{ ez_render_field( content, 'name' ) }}</h2>
+                        </div>
+
+                        {% if subFolders|length > 0 %}
+                            <ul class="break folder-view">
+                                {% for folder in subFolders %}
+                                    {% if folder.contentInfo.mainLocationId != location.contentInfo.mainLocationId %}
```
crevillo: what is this condition for? the list of subfolders could include the parent itself or something?

Helper/CriteriaHelper.php
```diff
@@ -21,6 +21,45 @@
...
+     * @return \eZ\Publish\API\Repository\Values\Content\Query\Criterion
+     */
+    public function generateSubContentCriterion( Location $location, array $languages = array() )
+    {
+        $criteria = array();
+        $criteria[] = new Criterion\Visibility( Criterion\Visibility::VISIBLE );
+        $criteria[] = new Criterion\Subtree( $location->pathString );
+        $criteria[] = new Criterion\ContentTypeIdentifier( array( 'folder' ) );
+        $criteria[] = new Criterion\LanguageCode( $languages );
+        $criteria[] = new Criterion\ParentLocationId( $location->id );
```
crevillo: do we need ParentLocationId criterion if we use Criterion\Subtree above?

Controller/DemoController.php
```diff
@@ -21,6 +21,130 @@
 class DemoController extends Controller
 {
     /**
+     * Displays the sub folder if it exists
+     *
+     * @param int $locationId of a folder
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function showSubFolderLinkAction( $locationId )
```
crevillo: ```SubFolderLinkAction```? Link makes me thing in a link to a subfolder instead of o a list of subfolders...

Controller/DemoController.php
```diff
@@ -21,6 +21,130 @@
...
+            $subFolders[] = $hit->valueObject;
+        }
+
+        return $this->render(
+            'eZDemoBundle:aside:folder.html.twig',
+            array(
+                'location' => $location,
+                'subFolders' => $subFolders
+            ),
+            $response
```
crevillo: as we have defined a custom controller for this, shouldn't we specificy ```SharedMaxAge``` here for caching?

Controller/DemoController.php
```diff
@@ -21,6 +21,130 @@
...
+            'eZDemoBundle:full:folder.html.twig',
+            array(
+                'location' => $location,
+                'content' => $content,
+                'pagerFolder' => $pager,
+                'subFolders' => $subFolders
+            ),
+            $response
+        );
+    }
```
crevillo: same comment as above for caching.

Controller/DemoController.php
clash82: Added.

Controller/DemoController.php
clash82: Added.

Controller/DemoController.php
andrerom: so subFolderListAction

crevillo: yep, that'd look better to me :)

Controller/DemoController.php
```diff
@@ -21,6 +21,134 @@
...
+     * Displays the sub folder if it exists
+     *
+     * @param int $locationId of a folder
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function showSubFolderLinkAction( $locationId )
+    {
+        $response = new Response();
+
+        $response->setSharedMaxAge( $this->getConfigResolver()->getParameter( 'content.default_ttl' ) );
```
crevillo: by default symfony makes this responses private, so dunno if we should add ```$response->setPublic()``` too. 
That said, dunno if the idea here was avoid cache. If not, shouldn't we set headers and Vary too?
https://github.com/ezsystems/DemoBundle/blob/master/Controller/DemoController.php#L80

Controller/DemoController.php
```diff
@@ -21,6 +21,134 @@
...
+            ->getContentService()
+            ->loadContentByContentInfo( $location->getContentInfo() );
+
+        // Getting language for the current siteaccess
+        $languages = $this->getConfigResolver()->getParameter( 'languages' );
+
+        // Using the criteria helper (a demobundle custom service) to generate our query's criteria.
+        // This is a good practice in order to have less code in your controller.
+        $criteria = $this->get( 'ezdemo.criteria_helper' )->generateListFolderCriterion(
+            $location, $this->container->getParameter( 'ezdemo.folder.folder_content.excluded_content_types' ), $languages
```
crevillo: won't this final ```,``` cause any problem?

crevillo: edit, sorry, i was worry about the last comma, but that was because i haven't seen the end of the line.

crevillo:
it seems there is a pull request related to this, but in this concrete one, we are deleting bootstrap.css but not updating page_head_style.html.twig accordingly. if we only merge this one we'll get an error with assetic generation.

Resources/public/less/layouts.less
```diff
@@ -155,7 +155,6 @@ aside {
     position: relative;
     > .subnavigation {
         > ul {
-            #gradient > .directional(fadeout(@grayLight, 70%), fadeout(@grayLight, 100%), -45deg);
```
andrerom: this change should be in the other PR

andrerom:
There is now some .css and .less changes here, I guess they belong to the other PR.

Otherwise +1

clash82:
@andrerom yes, all css/less mods belongs to #136

pspanja:
Location Search would be a better fit here (instead of Content Search), as the folder implies hierarchy and also all objects are rendered through sub-requests. This means Content Search will load full Content while there is no need for it.

clash82:
@pspanja just added, thanks :-)

pspanja:
@clash82 You should also use `$searchService->findContent( $subContentQuery );`, as this will still load Content, not Locations. Templates will need to be updated as in this case SearchHit will be a Location, not Content.

andrerom:
@pspanja did you mean `$searchService->findLocation( $subContentQuery )`

pspanja:
Yes, thanks :)

yannickroger:
@clash82 I am asking the same question once again :) 
> I think we could put this in a dedicated controller (at some point we'll have to refactor the demo controller, it does too much stuff).

andrerom:
@yannickroger Does not make sense to push that scope on him now, it is not part of the scope here.

andrerom:
@clash82 still css/less changes in this PR and very little time left for demo

clash82:
@pspanja thanks, just added...

clash82:
@andrerom css/less removed from commit.

andrerom:
+1, review ping @yannickroger @pspanja

yannickroger:
@andrerom I didn't mean the whole refactoring. Just create a dedicated controller like we have for Place, Feedback, languageswitcher etc…

andrerom:
Aha, any suggestions for a fitting controller for these two? (_naming things_)

Resources/config/ezdemo.yml
```diff
@@ -59,7 +64,11 @@ system:
                     template: "eZDemoBundle:full:call_to_action.html.twig"
                     match:
                         Identifier\ContentType: "call_to_action"
-
+                folder:
+                    controller: "eZDemoBundle:Demo:showFolderList"
+                    template: "eZDemoBundle:full:folder.html.twig"
+                    match:
+                        Identifier\ContentType: "folder"
```
yannickroger: missing blank line ?

clash82:
I've added separate controllers, any suggestions?

andrerom:
No, good for me but @yannickroger had a small one above.

yannickroger:
+1
But we need to add BDD as a follow up or have QA do it while testing it.

Helper/CriteriaHelper.php
```diff
@@ -21,6 +21,46 @@
...
+     * @param string[] $excludeContentTypeIdentifiers Array of excluded contentType identifiers
+     * @param string[] $languages Array of languages
+     *
+     * @return \eZ\Publish\API\Repository\Values\Content\Query\Criterion
+     */
+    public function generateListFolderCriterion( Location $location, array $excludeContentTypeIdentifiers = array(), array $languages = array() )
+    {
+        $criteria = array();
+        $criteria[] = new Criterion\Visibility( Criterion\Visibility::VISIBLE );
+        $criteria[] = new Criterion\Subtree( $location->pathString );
```
pspanja: Using `Subtree` instead of `ParentLocationId` will fetch from the whole subtree, I gather this is not intended?

crevillo: Yep, seems kinda redundant to me. I'd go for ParentLocationId too

Resources/views/aside/folder.html.twig
```diff
@@ -0,0 +1,12 @@
+{% if treeChildItems|length > 0 %}
+    <ul>
+        {% for treeChildItem in treeChildItems %}
+             <li><a href="{{ path( 'ez_urlalias', { 'locationId': treeChildItem.contentInfo.mainLocationId } ) }}" title="{{ ez_content_name( treeChildItem.contentInfo ) }}">{{ ez_content_name( treeChildItem.contentInfo ) }}</a></li>
+        {% endfor %}
+    </ul>
+{% elseif isChild is not defined %}
```
pspanja: AFAICS this logic will render two levels when used through folder full view, as the sub-request from sub-navigation will enter here, and sub-request below will have `isChild` set to true? I kind of find it hard to read.

Resources/views/full/folder.html.twig
```diff
@@ -0,0 +1,43 @@
...
+
+                </div>
+
+                <aside class="col-md-4">
+                    <section class="subnavigation row nav-collapse">
+                        <div class="attribute-header">
+                            <h2>{{ ez_render_field( content, 'name' ) }}</h2>
+                        </div>
+
+                        <ul class="break folder-view">
```
pspanja: It would be better to check if the list is empty and only then render `<ul>` tags, otherwise you'll render it empty in that case.
Also it seems the sub-request will render `<ul><ul>`, or am I missing something?

Resources/views/aside/folder.html.twig
```diff
@@ -0,0 +1,12 @@
...
+    <li>
+        <a href="{{ path( 'ez_urlalias', { 'locationId': location.contentInfo.mainLocationId } ) }}" title="{{ ez_content_name( location.contentInfo ) }}">{{ ez_content_name( location.contentInfo ) }}</a>
+        {{ render_esi( controller( 'eZDemoBundle:Folder:showFolderListAsideView', { 'locationId': location.contentInfo.mainLocationId } ) ) }}
```
pspanja: IMHO it is better practice to use `ez_content:viewLocation` and configure controller with the view.
The template could then be general as it renders only linked name, so nothing folder specific.

Resources/views/aside/folder.html.twig
clash82: Exactly as you think :-)

Resources/views/aside/folder.html.twig
```diff
@@ -0,0 +1,12 @@
+{% if treeChildItems|length > 0 %}
```
bdunogier: One readability nitpick: `if treeChildItems is not empty` is more readable imho.

Resources/views/aside/folder.html.twig
bdunogier: +1 with @pspanja.

Resources/views/full/folder.html.twig
```diff
@@ -0,0 +1,45 @@
...
+
+                </div>
+
+                <aside class="col-md-4">
+                    <section class="subnavigation row nav-collapse">
+                        <div class="attribute-header">
+                            <h2>{{ ez_render_field( content, 'name' ) }}</h2>
+                        </div>
+
+                        {% if treeItems|length > 0 %}
```
bdunogier: Same remark as above, `if treeItems is not empty` would be more readable.

Resources/config/ezdemo.yml
```diff
@@ -78,6 +89,11 @@ system:
                     template: "eZDemoBundle:line:line.html.twig"
                     match:
                         Identifier\ContentType: [landing_page, call_to_action, product, blog]
+                folder:
+                    template: "eZDemoBundle:line:folder.html.twig"
+                    match:
+                        Identifier\ContentType: "[folder]"
```
bdunogier: You don't need the quotes on arrays.

clash82: Got it, thanks :-)

Resources/views/full/folder.html.twig
clash82: You are right, looks better now.

Controller/FolderController.php
```diff
@@ -0,0 +1,162 @@
...
+
+class FolderController extends Controller
+{
+    /**
+     * Displays the sub folder if it exists
+     *
+     * @param int $locationId of a folder
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function showFolderListAsideViewAction( $locationId )
```
bdunogier: Since @yannickroger has merged the ParamConverter PR, you can directly typehint the Location object, and it will be loaded for you:

```
public function showFolderListAsideViewAction( Location $location )
```
And of course, you can remiove the getRepository, getLocationService and loadLocation calls. below (but keep the invisible check).

The same improvement can be applied to all controller actions, as long as the argument is a Location or a Content.

crevillo: off topic: always wondering about why we are using NotFoundHttpException for invisible locations...

Controller/FolderController.php
```diff
@@ -0,0 +1,162 @@
+<?php
+/**
+ * File containing the FolderController class.
```
bdunogier: We no longer add the class name in the header (it's already there in the class anyway). Use "This file is part of the ezsystems/demo-bundle package".

Resources/config/default_settings.yml
```diff
@@ -8,6 +8,12 @@ parameters:
     ezdemo.footer.latest_content.content_types: [article]
     # Number of blog_posts displayed per page in the blog view
     ezdemo.blog.blog_post_list.limit: 10
+    # Number of article displayed per page in the folder view
+    ezdemo.folder.folder_list.limit: 10
+    # ContentType identifiers we want to exclude from content list in folder view
+    ezdemo.folder.folder_content.excluded_content_types: [folder, image]
```
bdunogier: I'd have used `folder_view.excluded_content_types` instead, as it is the list of "content types excluded from the folder view".

Resources/views/full/folder.html.twig
```diff
@@ -0,0 +1,45 @@
+{% extends "eZDemoBundle::pagelayout.html.twig" %}
+
+{% block content %}
+    <section class="content-view-full">
+        <div class="class-folder">
+            <div class="row">
+                <div class="col-md-8">
+                    <div class="attribute-header">
+                        <h1>{{  ez_render_field( content, 'name' ) }}</h1>
```
crevillo: nitpick: due this is our demobundle i will maybe use ```{{ ez_content_name (content) }}``` just to let the newcomers know about it.

Controller/FolderController.php
clash82: @yannickroger and what about $location? should I change this:

`if ( $location->invisible )`
to this:
`if ( $locationId->invisible )`
or?

Controller/FolderController.php
```diff
@@ -0,0 +1,159 @@
...
+
+class FolderController extends Controller
+{
+    /**
+     * Displays the sub folder if it exists
+     *
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location of a folder
+     * @return \Symfony\Component\HttpFoundation\Response
+     */
+    public function showFolderListAsideViewAction( Location $location )
```
lolautruche: This action doesn't seem to be called according to your configuration.

Controller/FolderController.php
```diff
@@ -0,0 +1,159 @@
...
+        $subContent = $searchService->findLocations( $subContentQuery );
+
+        $treeItems = array();
+        foreach ( $subContent->searchHits as $hit )
+        {
+            $treeItems[] = $hit->valueObject;
+        }
+
+        return $this->render(
+            'eZDemoBundle:full:folder.html.twig',
```
lolautruche: Please avoid hardcoding the template when using a custom controller. Moreover, you already have specified your template in the config.

What you should just do is:

```php
return $this->get( 'ez_content' )->viewLocation(
    $location->id,
    'full',
    true,
    ['pageFolder' => $pager, 'treeItems' => $treeItems]
);
```

lolautruche: Reason is that this will avoid issues with preview for instance, as template selection will be based on the siteaccess settings.

Note that the remark is also valid for the action above this one.

Controller/FolderController.php
```diff
@@ -0,0 +1,159 @@
...
+use eZ\Publish\API\Repository\Values\Content\Query;
+use eZ\Publish\API\Repository\Values\Content\Query\SortClause;
+use eZ\Publish\API\Repository\Values\Content\Location;
+
+class FolderController extends Controller
+{
+    /**
+     * Displays the sub folder if it exists
+     *
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location of a folder
```
bdunogier: Don't forget to add an @throws for the `NotFoundHttpException` below.

bdunogier:
And please, don't rebase the whole branch into one commit everytime. It makes it easier to follow the conversation if commits are more granular.

Rebasing is good practice, but after things have settled down a bit :-)

andrerom:
Yes, we where a bit unclear when we mentioned you should rebase earlier. What we specifically try to do is that we tend to rebase _before merge_ (so after review has passed) into one commit, or in case where PR solves several things, into logical commits.

@clash82 This is especially important next month as we then have another person on your team, and then constant rebasing makes it very hard to work together on a PR ;)

Controller/FolderController.php
```diff
@@ -0,0 +1,159 @@
...
+            new SortClause\DatePublished()
+        );
+
+        // Initialize pagination.
+        $pager = new Pagerfanta(
+            new ContentSearchAdapter( $query, $this->getRepository()->getSearchService() )
+        );
+
+        $pager->setMaxPerPage( $this->container->getParameter( 'ezdemo.folder.folder_list.limit' ) );
+        $pager->setCurrentPage( $this->getRequest()->get( 'page', 1 ) );
```
wizhippo: Should add Request $request to the function parameters as getRequest is deprecated since version 2.4 of symfony

wizhippo:
ezp4 also included rss feed in it's folders.  I know ezdemo is not a replacement for all of ezp4 and this is probably out of scope for this commit.  Just though I would point it out.

Controller/FolderController.php
lolautruche: Yes, you're right @wizhippo

clash82:
> Yes, we where a bit unclear when we mentioned you should rebase earlier. What we specifically try to do is that we tend to rebase before merge (so after review has passed) into one commit, or in case where PR solves several things, into logical commits.

@andrerom got it! rebase only after final review :-)

Controller/FolderController.php
clash82: @lolautruche can you be more specific? I'm not quite sure if I understand.

clash82:
I'm not quite sure if `aside_sub` is a good name. @pspanja @bdunogier review for last commit appreciated.

Controller/FolderController.php
bdunogier: If I recall correctly, this controller wasn't referenced anywhere.

What is it called by ?

bdunogier:
Looks much better with the latest changes.

andrerom:
`aside_sub` is ok, and you actualy have two +1's so ready to (optionally rebase and) merge

clash82:
Great :-) Rebased and waiting for merge.

Controller/FolderController.php
andrerom: fixed on line below

andrerom:
Done, thanks for the effort, first one takes some time, with time it only gets easier* :)

* lots of conventions to learn for the first few pull requests

clash82:
Wow, thank you for all precious tips. I've learned lots of new things. It should go faster (and smarter) next time :-)

> BTW Remember to add JIRA bug/improvement/story for next one so it becomes part of changelog ;)

Yes Sir! ;-)

PR #139, opened 2015-02-10
Fixed remaining legacy dependencies

bdunogier:


andrerom:
+1

Resources/views/line/blog_post.html.twig
```diff
@@ -1,7 +1,7 @@
 {# In order to use blocks and override ez_render_field's template within this file,
  # we need to extend from another template and surround our content with a block.
  #}
-{% extends viewbaseLayout %}
+{% extends 'EzPublishCoreBundle::viewbase_layout.html.twig' %}
```
crevillo: Ok, but why this is related to decoupling legacy?

composer.json
```diff
@@ -15,7 +15,8 @@
         "ezsystems/comments-bundle": "@dev",
         "ezsystems/platform-ui-assets-bundle": "~0.1",
         "whiteoctober/breadcrumbs-bundle": "~1.0.1",
-        "knplabs/knp-menu-bundle": "1.*"
+        "knplabs/knp-menu-bundle": "1.*",
+        "zetacomponents/feed": "~1.4"
```
crevillo: Hmmm, why? Do we need or will need it?

crevillo:
My comments are just curiosity, +1 anyway.

composer.json
lolautruche: Yes it's needed for the feed block on the home page.

lolautruche:
+1

Resources/views/line/blog_post.html.twig
andrerom: @lolautruche this (afaik) is the workaround talked about in todays daily.

bdunogier: It is. And I can confirm that it still occurs if I revert this patch.

Trying to investigate.

andrerom: @dpobel is this the one you mean? `{% extends noLayout ? viewbaseLayout : "eZDemoBundle::pagelayout.html.twig" %}`

PR #140, opened 2015-02-13
Fix unused response in FolderController

wizhippo:
The response from viewLocation should be used as the the response created is not used.  This also adds support for different view types.

### DIFF
--- a/Controller/FolderController.php
+++ b/Controller/FolderController.php
@@ -29,7 +29,7 @@ class FolderController extends Controller
      * @throws NotFoundHttpException $location is flagged as invisible
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function showFolderListAsideViewAction( Location $location, $viewType, $layout = false )
+    public function showFolderListAsideViewAction( Location $location, $viewType, $layout = false, array $params = array() )
     {
         if ( $location->invisible )
         {
@@ -64,7 +64,7 @@ class FolderController extends Controller
             $location->id,
             $viewType,
             $layout,
-            ['treeChildItems' => $treeChildItems]
+            array( 'treeChildItems' => $treeChildItems ) + $params
         );
     }
 
@@ -76,7 +76,7 @@ class FolderController extends Controller
      * @throws NotFoundHttpException $location is flagged as invisible
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function showFolderListAction( Request $request, Location $location, $viewType, $layout = false )
+    public function showFolderListAction( Request $request, Location $location, $viewType, $layout = false, array $params = array() )
     {
         if ( $location->invisible )
         {
@@ -139,7 +139,7 @@ class FolderController extends Controller
             $location->id,
             $viewType,
             $layout,
-            ['pagerFolder' => $pager, 'treeItems' => $treeItems]
+            array( 'pagerFolder' => $pager, 'treeItems' => $treeItems ) + $params
         );
     }
 }
--- a/Controller/FolderController.php
+++ b/Controller/FolderController.php
@@ -29,18 +29,13 @@ class FolderController extends Controller
      * @throws NotFoundHttpException $location is flagged as invisible
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function showFolderListAsideViewAction( Location $location )
+    public function showFolderListAsideViewAction( Location $location, $viewType, $layout = false )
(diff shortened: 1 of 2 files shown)


Controller/FolderController.php
```diff
@@ -65,12 +60,18 @@ public function showFolderListAsideViewAction( Location $location )
...
+        $response = $this->get( 'ez_content' )->viewLocation(
             $location->id,
-            'aside_sub',
-            true,
+            $viewType,
+            $layout,
             ['treeChildItems' => $treeChildItems]
         );
+
+        $response->setSharedMaxAge( $this->getConfigResolver()->getParameter( 'content.default_ttl' ) );
```
crevillo: If i'm not wrong this is not needed as ViewLocation do the Same just before.

Controller/FolderController.php
```diff
@@ -146,11 +141,17 @@ public function showFolderListAction( Request $request, Location $location )
...
-        return $this->get( 'ez_content' )->viewLocation(
+        $response = $this->get( 'ez_content' )->viewLocation(
             $location->id,
-            'full',
-            true,
+            $viewType,
+            $layout,
             ['pagerFolder' => $pager, 'treeItems' => $treeItems]
         );
+
```
crevillo: Same as above

Controller/FolderController.php
wizhippo: You are correct.  It also sets X-Location-Id

andrerom:
Review ping @yannickroger @dpobel

Controller/FolderController.php
```diff
@@ -29,19 +29,13 @@ class FolderController extends Controller
...
+    public function showFolderListAsideViewAction( Location $location, $viewType, $layout = false, array $params = array() )
     {
-        $response = new Response();
-
         if ( $location->invisible )
         {
             throw new NotFoundHttpException( "Location #$location->id cannot be displayed as it is flagged as invisible." );
         }
 
-        $response->setSharedMaxAge( $this->getConfigResolver()->getParameter( 'content.default_ttl' ) );
```
clash82: We don't cache it anymore?

crevillo: it will be cached because now we are using ``` $this->get( 'ez_content' )->viewLocation```, and that will take care of it. 
if we don't use viewLocation as it is here know then we will need to add cache directives.

crevillo: actually,  @wizhippo and all, do we need the location->invisible check now we're calling viewLocation?

clash82: Understood, thanks :-)

wizhippo: @crevillo Your are right.  I have removed this too.

clash82: @wizhippo in this case you should also remove
`@throws NotFoundHttpException $location is flagged as invisible`

Controller/FolderController.php
```diff
@@ -29,19 +29,8 @@ class FolderController extends Controller
      * @throws NotFoundHttpException $location is flagged as invisible
```
andrerom: I think @clash82 suggested this can be removed as well now, same with the other method and then also the `use` line at the top.

clash82: :+1:

Controller/FolderController.php
```diff
@@ -26,22 +25,10 @@ class FolderController extends Controller
      * Displays the sub folder if it exists
      *
      * @param \eZ\Publish\API\Repository\Values\Content\Location $location of a folder
-     * @throws NotFoundHttpException $location is flagged as invisible
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function showFolderListAsideViewAction( Location $location )
+    public function showFolderListAsideViewAction( Location $location, $viewType, $layout = false, array $params = array() )
```
andrerom: Some questions:
Now this name is not really correct anymore is it?
Is it so general the two methods could be merged?
and where is 'aside' specified now btw, or?

wizhippo: I agree they probably can be combined. @clash82  perhaps knows better why the aside was split  how it is.

clash82: `showFolderListAsideViewAction` is used to display second level items in folders tree. We don't want to have recursive actions, all we need is to display only first level subfolders.

You may combine those two actions but be aware to assign some variable that will inform second level request to exit instead of making more (and more...) recursive requests...

PR #141, opened 2015-02-19
EZP-24015: Improve Language Switcher flags and logic

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-24015

This PR requires:
- [x] https://github.com/ezsystems/PlatformUIAssetsBundle/pull/3
- [x] #136 

to be merged first.


clash82:
Waiting for review.
ping @sunpietro @bdunogier @dpobel @lolautruche @crevillo @lserwatka @andrerom

Resources/views/page_head_style.html.twig
```diff
@@ -5,6 +5,8 @@
     <link rel="stylesheet" type="text/css" href="{{ asset_url }}"/>
 {% endstylesheets %}
 
+<link rel="stylesheet" type="text/css" href="/bundles/ezplatformuiassets/vendors/flag-icon-css/css/flag-icon.min.css" />
```
dpobel: you should add this stylesheet in the list above instead (with the `@Bundle` notation)

Resources/views/parts/languages_switcher.html.twig
```diff
@@ -4,19 +4,29 @@
...
+            {% for lang in availableLanguages %}
+                {% do routeRef.set( "siteaccess", ezpublish.translationSiteAccess( lang ) ) %}
+                    {% if lang == currentLang %}
+                        <li class="current">
+                            <div class="flag-icon flag-icon-background flag-icon-{{ lang[4:]|lower }}"></div>{{ lang[:3]|lower|capitalize }}
+                        </li>
+                    {% else %}
+                        <li>
+                            <a href="{{ url( routeRef ) }}">
+                                <div class="flag-icon flag-icon-background flag-icon-{{ lang[4:]|lower }}"></div>{{ lang[:3]|lower|capitalize }}
```
dpobel: You should use span instead of a div

Resources/views/parts/languages_switcher.html.twig
```diff
@@ -4,19 +4,29 @@
...
 #    - 'availableLanguages'
 #}
-<div id="lang-selector">
-    {# Looping over all available languages to display the links #}
 
-    {% for lang in availableLanguages %}
-        {# This time, we alter the "siteaccess" parameter directly. #}
-        {# We get the right siteaccess with the help of ezpublish.translationSiteAccess() helper #}
+{% if availableLanguages|length > 1 %}
+    <div id="lang-selector" class="transition-showed">
```
sunpietro: When the class ``transition-showed`` is used? BTW. the class name is incorrect.
Do you really need ID there?

Resources/views/parts/languages_switcher.html.twig
```diff
@@ -4,19 +4,29 @@
...
-                {{ siteaccess[lang]}}
-            </a>
-        {% endif %}
-    {% endfor %}
-</div>
+            {% for lang in availableLanguages %}
+                {% do routeRef.set( "siteaccess", ezpublish.translationSiteAccess( lang ) ) %}
+                    {% if lang == currentLang %}
+                        <li class="current">
+                            <div class="flag-icon flag-icon-background flag-icon-{{ lang[4:]|lower }}"></div>{{ lang[:3]|lower|capitalize }}
```
sunpietro: I would use `span` there instead of `div`

Resources/views/parts/languages_switcher.html.twig
```diff
@@ -4,19 +4,29 @@
...
+                            <a href="{{ url( routeRef ) }}">
+                                <div class="flag-icon flag-icon-background flag-icon-{{ lang[4:]|lower }}"></div>{{ lang[:3]|lower|capitalize }}
+                            </a>
+                        </li>
+                    {% endif %}
+            {% endfor %}
+        </ul>
+
+        <a href="#lang-selector" class="current-lang">
+            <div class="current-language flag-icon flag-icon-background flag-icon-{{ currentLanguage[4:]|lower }}"></div>{{ currentLanguage[:3]|lower|capitalize }}&nbsp;↴
```
sunpietro: Block elements shouldn't be placed inside inline elements.

Resources/public/css/bootstrap.css
```diff
@@ -8942,7 +8942,7 @@ del:hover {
 }
 .navbar.extra-navi #lang-selector a {
   display: inline-block;
-  padding: 5px 10px 5px 30px;
+  padding: 5px 10px 5px 10px;
```
sunpietro: Nitpick: You can write like this: `padding: 5px 10px;`

Resources/views/page_head_style.html.twig
clash82: It would be nice, but i can't. CSS contains relative path for *.svg dependencies. Assetic puts CSS away from original location.

dpobel: As far as I can remember, Assetic has a filter for that but indeed, it had some issues in the past, did you test it ?

Resources/views/parts/languages_switcher.html.twig
clash82: ID's are cool... sometimes. This one is used with CSS :target element.

Resources/views/page_head_style.html.twig
clash82: There is a `cssrewrite` filter for assetic, but it still requires second request to sever (paths must be corrected only for flags styles)... besides, it doesn't work as it should. I don't know how to solve it right now.

andrerom: the second request to server point is only needed for dev env, in prod it will be done when you dump assets afaik. however if it does not work that is something else.

clash82: It doesn't work. After rewriting paths to *.svg, they still refers to wrong (different) place.

@andrerom Second request will also occur in prod env. It's because there will be second <link> tag. `cssrewrite` must be used only for one css file, so I must put it into separate `{% stylesheet %}` block. Assetic doesn't have option to use filter only for one file in entire block.

andrerom: ah, that is what you meant. ok, but that is already the case with the code above it seems as it is separate link tag(?)

clash82: Any of these cases will generate separate link tag. There's no other way.

andrerom:
+1

sunpietro:
+0.99

clash82:
Candidate for merge?
ping @andrerom

Resources/views/page_head_style.html.twig
kamilmusial: **@eZPlatformUIAssetsBundle** syntax causes path mismatch and flags cannot be displayed properly, [see link](http://symfony.com/doc/current/cookbook/assetic/asset_management.html#including-css-stylesheets).
Besides that, **bootstrap** isn't able to reach its font for _glyphicons_ (`../fonts/glyphicons-halflings-regular.woff2` - `404`) And afaik never was.
I think there are two possible ways to fix it:
1. Leave unused Bootstrap Glyphicons feature as unused and set cssrewrite filter. Also change syntax to:
```php
{% stylesheets filter="cssrewrite"
    'bundles/ezdemo/css/video.css'
    'bundles/ezdemo/css/bootstrap.css'
    'bundles/ezplatformuiassets/vendors/flag-icon-css/css/flag-icon.min.css'
%}
```
Then we have one asset and working flags.
2. Copy Bootstrap core into DemoBundle
Then we can set output for **cssrewrite** and try something [like this](https://coderwall.com/p/cx1ztw/bootstrap-3-in-symfony2-with-no-extra-bundles)

andrerom: Would prefer 1., as we are trying to avoid to copy external libraries into our repos (git history size, separation, reuse).

clash82: If no more comments I'm gonna apply solution proposed by Kamil (1) tomorrow.

andrerom: @dpobel is not here today, but check with @lserwatka and go ahead if he agrees :)

clash82: Unfortunately @lserwatka will be at work on Monday so I guess we will just have to wait.

andrerom: then lets rather ask @dpobel, he should be back today :)

clash82: @dpobel give us some feedback :-)

dpobel: +1 on using `'bundles/ezplatformuiassets/vendors/flag-icon-css/css/flag-icon.min.css'`, this is the recommended way in the Symfony documentation: http://symfony.com/doc/current/cookbook/assetic/asset_management.html#fixing-css-paths-with-the-cssrewrite-filter

Side note: at some point, we should try to investigate that issue, it's really a pity to have such limitation...

clash82:
I've added changes proposed by @kamilmusial. Please, take a look on last commit.
ping @andrerom @dpobel @lserwatka

lserwatka:
We spent a lot of effort to externalize Bootstrap, so I second first solution from Kamil. 
+1

andrerom:
Merged, thanks @clash82  and @kamilmusial !

clash82:
tu :-)

PR #144, opened 2015-02-24
[WIP] Change use of travis to cleanly support custom branches(and forks)

andrerom:
This is work in progress, should not be reviewed before passing on travis

Issue: travis fails many times because of the way we inject repo into ezpublish-community checkout, typically issue on asset changes, but also on dependency changes.

Plan: add [repo reference](https://getcomposer.org/doc/04-schema.md#repositories) to local checkout and set requirement to travis branch using [env variables](http://docs.travis-ci.com/user/ci-environment/#Environment-variables) before running composer install, try to extract this into script we can put in ezpublish-community.

- [ ] Reproduce the issue cause by the way behat testing is currently done (assets, composer deps issues)
- [ ] Apply fix (see https://github.com/ezsystems/ezpublish-kernel/commit/7223c3d0bbeeb592b686794f65c05af57dec10e9)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ php:
 # list of paths/bundle to execute
 env:
     # test demo with content
-  - INSTALL="demoContentNonUniqueDB" PROFILE="demo" TEST="content"
+#  - INSTALL="demoContentNonUniqueDB" PROFILE="demo" TEST="content"
     # test demo w/o content
   - PROFILE="demo" TEST="clean"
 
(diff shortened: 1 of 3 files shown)


andrerom:
This pr is supposed to have 2 commits not present in master, and the commit above was supposed to fail. something strange here.

andrerom:
(the revert is only to have something to test this on, aka make it fail first, then make it pass)

andrerom:
@clash82 Any suggestions on how to best reproduce the issue you had in boostrap PR?

clash82:
Good question. Templates are not compiling properly... but why?

andrerom:
Might have fixed it in #150, closing.

PR #147, opened 2015-03-20
Fix for EZP-24015: Improve Language Switcher flags and logic

pedroresende:
### Fix for EZP-24015: Improve Language Switcher flags and logic

After the [EZP-24015: Improve Language Switcher flags and logic](https://github.com/ezsystems/DemoBundle/pull/141/files) a typo has been introduced, instead of currentLanguage it was currentLang.


clash82:
Yes, It's my fault. Should be `currentLanguage`. Sorry for that.
:+1:

andrerom:
+1

PR #148, opened 2015-03-31
Fix EZP-24052: Preview causes redirect to versionview not being able to continue editing

lolautruche:
> https://jira.ez.no/browse/EZP-24052

See https://github.com/ezsystems/LegacyBridge/pull/23

### DIFF
--- a/Resources/views/pagelayout.html.twig
+++ b/Resources/views/pagelayout.html.twig
@@ -21,8 +21,8 @@
 {% endif %}
 
 <!-- Complete page area: START -->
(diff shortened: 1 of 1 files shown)


lolautruche:
Review ping @andrerom @bdunogier @dpobel @glye @pspanja @yannickroger

andrerom:
won't this break when not having legacy bridge installed?

lolautruche:
You're right, fixed :blush:

yannickroger:
+1

andrerom:
+1

PR #151, opened 2015-04-17
[WIP] EZP-24231: As a Developer I would like to have more abstract structure of the PrivacyCookieBundle

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-24231

**FYI, I want to rework this PR. Right now I don't have time for it.**


clash82:
ping @andrerom @bdunogier @michalpipa

DependencyInjection/eZDemoCompilerPass.php
```diff
@@ -0,0 +1,39 @@
...
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\DemoBundle\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\Reference;
+
+class eZDemoCompilerPass implements CompilerPassInterface
```
bdunogier: Why do you need this Compiler Pass ?

bdunogier:
This sounds a bit complicated, imho. I think you should start with a first version that adds the required configuration, andconsider the content based one for later.

DependencyInjection/eZDemoCompilerPass.php
clash82: Currently I don't know any better idea to plug into PrivacyCookieBundle and override existing banner settings.

PrivacyCookie/DemoBundleBanner.php
```diff
@@ -0,0 +1,36 @@
...
+ * Builds a Privacy Cookie Banner object based on a configured Content Object.
+ */
+class DemoBundleBanner implements DemoBundleBannerFactory
+{
+    public static function build( Content $bannerContent, LocationService $locationService, URLAliasService $urlAliasService, Translator $translator )
+    {
+        $banner = new Banner();
+
+        $banner->caption = $bannerContent->getFieldValue( 'description' )->xml->textContent;
+        $banner->policyPageUrl = $urlAliasService->reverseLookup(
```
crevillo: a question here. is the goal is to make this bundle usable even from not ez publish installations shouldn't we avoid using ez5 services?

clash82: This is DemoBundle not a PrivacyCookieBundle.

crevillo: yes, sorry @clash82, i missread.

clash82: np, I also love Mondays ;-)

DependencyInjection/eZDemoCompilerPass.php
bdunogier: The easiest way is to use service aliases, usually.

In the PrivacyCookieBundle, the factory service, `ez_privacy_cookie.banner_factory`, is an alias for `ez_privacy_cookie.configuration_banner_factory`. In the DemoBundle, just redeclare this service to be an alias of yours.

And then you should not need the compiler pass anymore.

PrivacyCookie/DemoBundleBanner.php
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Builds a Privacy Cookie Banner object based on a configured Content Object.
+ */
+class DemoBundleBanner implements DemoBundleBannerFactory
+{
+    public static function build( Content $bannerContent, LocationService $locationService, URLAliasService $urlAliasService, Translator $translator )
+    {
+        $banner = new Banner();
+
+        $banner->caption = $bannerContent->getFieldValue( 'description' )->xml->textContent;
```
crevillo: question here. why we use ```->xml->textContent``` here? If description is a xmlblock and the user wants some bold text or whatever, won't will loose them?

bdunogier: >
> > +        $banner->caption = $bannerContent->getFieldValue( 'description' )->xml->textContent;
>
> question here. why we use ->xml->textContent here? If description is a
> xmlblock and the user wants some bold text or whatever, won't will loose
> them?
>
Carlos is right. I'm not sure we should aim at supporting rich text for
caption straight away. It can be an improvement.

crevillo: not sure either, @bdunogier. If we let text for caption, maybe some user may thing in adding another image as caption... i will go for a textlinea or textarea. 

bonus topic for platform ui -> it would be awesome if we could define some kinda "rich text profiles" and devs could decide which rich text profile apply to a concrete field. i mean, maybe for a field you wan to allow bolds, italic but don't want bullets or images...

clash82: IMO plain text is the best option here. Message banner is not intended to display rich text or images - it should be treated as a later improvement.

bdunogier: +1 on the "later" part :-)

andrerom:
Didn't quite get why there are changes to the demo data here, could that be explained?

clash82:
> Didn't quite get why there are changes to the demo data here, could that be explained?

New article content was added to allow displaying privacy policy banner. This is the location where user will be redirected after clicking `learn more` anchor. Besides, article `title` and `description` text is shown on the notification banner - user can easily personalize privacy banner without need to edit templates or configuration files.

DependencyInjection/eZDemoCompilerPass.php
```diff
@@ -0,0 +1,39 @@
...
+
+        $definition = $this->container->getDefinition( 'ez_privacy_cookie.twig.extension' );
+        $eZBannerFactory = $this->container->getDefinition( 'ezdemo.ez_content_banner_factory' );
+        $definition->replaceArgument( 2, $eZBannerFactory );
+    }
+
+    public function process( ContainerBuilder $container )
+    {
+        $this->container = $container;
+
```
crevillo: What's is this for? I mean, not used in the function, so...

clash82: Hmmm, but `$this->container` is used by `addPrivacyCookieBanner()` method?

I'm going to try remove compiler pass anyway.

crevillo: yep, is used there. just wondering if we could maybe merge the two functions in an only one... 
but leave that to @andrerom @bdunogier .

DependencyInjection/eZDemoCompilerPass.php
```diff
@@ -0,0 +1,39 @@
...
+        $this->addPrivacyCookieBanner();
```
bdunogier: Why don't you just add move the code from addPrivacyCookieBanner ?

We usually have Compiler passes that do one thing each. This one adds the privacy cookie banner, no need to create a private method right ?

clash82: ok, now I get it, thanks :-)

clash82:
I'd like to get your opinion about this PR? 
Can we merge it or close it?

ping @andrerom @bdunogier

DependencyInjection/eZDemoCompilerPass.php
```diff
@@ -0,0 +1,29 @@
...
+            return;
+        }
+
+        if ( $container->hasDefinition( 'ezdemo.ez_content_banner_factory' ) === false )
+        {
+            return;
+        }
+
+        $definition = $container->getDefinition( 'ez_privacy_cookie.twig.extension' );
+        $eZBannerFactory = $container->getDefinition( 'ezdemo.ez_content_banner_factory' );
```
bdunogier: `$eZ...` won't work with coding standards. What about `$definition->replaceArgument( 1, $container->getDefinition( 'ezdemo.ez_content_banner_factory' ) );` ?

DependencyInjection/eZDemoCompilerPass.php
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * @license For full copyright and license information view LICENSE file distributed with this source code.
+ */
+
+namespace EzSystems\DemoBundle\DependencyInjection;
+
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+
+class eZDemoCompilerPass implements CompilerPassInterface
```
bdunogier: I'd name it after what it does, not after what it belongs to.

`PrivacyCookiePass` ?

clash82:
+1 :)

DependencyInjection/PrivacyCookiePass.php
```diff
@@ -0,0 +1,28 @@
...
+            return;
+        }
+
+        if ( $container->hasDefinition( 'ezdemo.ez_content_banner_factory' ) === false )
+        {
+            return;
+        }
+
+        $definition = $container->getDefinition( 'ez_privacy_cookie.twig.extension' );
+        $definition->replaceArgument( 1, $container->getDefinition( 'ezdemo.ez_content_banner_factory' ) );
```
bdunogier: I wouldn't use a compiler pass for this. There is a service alias that is meant for this exactly:

```yaml
ez_privacy_cookie.banner_factory:
    class: %ez_privacy_cookie.banner_factory.class%
    alias: ez_privacy_cookie.configuration_banner_factory
```

Just redefine this alias in the DemoBundle's `service.yml` file:
```yaml
ez_privacy_cookie.banner_factory:
    class: %ez_privacy_cookie.banner_factory.class%
    alias: ezdemo.ez_content_banner_factory
```

PrivacyCookie/DemoBundleBannerFactory.php
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+namespace EzSystems\DemoBundle\PrivacyCookie;
+
+use eZ\Publish\API\Repository\LocationService;
+use eZ\Publish\API\Repository\URLAliasService;
+use eZ\Publish\API\Repository\ContentService;
+use Symfony\Component\Translation\Translator;
+
+interface DemoBundleBannerFactory
```
bdunogier: Why do you declare a new BannerFactory interface ? You already have one in the privacy cookie bundle package: https://github.com/ezsystems/EzSystemsPrivacyCookieBundle/blob/master/Twig/PrivacyCookieTwigExtension.php#L29.

PrivacyCookie/DemoBundleBanner.php
```diff
@@ -0,0 +1,74 @@
...
+use eZ\Publish\API\Repository\ContentService;
+use eZ\Publish\API\Repository\URLAliasService;
+use EzSystems\PrivacyCookieBundle\Banner\Banner;
+use EzSystems\DemoBundle\PrivacyCookie;
+use Symfony\Component\Translation\Translator;
+
+/**
+ * Builds a Privacy Cookie Banner object based on a configured Content Object.
+ */
+class DemoBundleBanner implements DemoBundleBannerFactory
```
bdunogier: As said in another comment, the Factory must implement the privacy-cookie-bundle's interface, not a local one.

PrivacyCookie/DemoBundleBanner.php
```diff
@@ -0,0 +1,74 @@
...
+{
+    public static function build( array $bannerOptions, LocationService $locationService, ContentService $contentService, URLAliasService $urlAliasService, Translator $translator )
```
bdunogier: The services used by the factory must be passed to the constructor, not to the build method. The build method is required by the implement bundle's (privacy-cookie-bundle) interface.

PrivacyCookie/DemoBundleBanner.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        $contentId = null;
+
+        // check if `Content` object path exists and if so, get associated `contentId`
```
bdunogier: We usually favour `contentId` over `pathString`. It would also avoid the alias lookup below.

bdunogier:
Frankly, I'm not convinced.

For one part because of the PrivacyCookieBundle itself, where the Factory isn't really used as one. I guess i've missed it during reviews.

I'd have expected a simpler pattern: the twig extension requires the BannerFactory (injected). The factory service has an alias that we can replace in our configuration, here in the demobundle. When the twig extension is called, the BannerFactory's `build()` method is called, and we use the obtained Banner object.

I also find the Factory to be quite complicated here. I've made a few suggestions that could make it more readable. The support of both configuration & Content data is imho a bit too much in one place, but I can live with that. You could look into using Symfony's OptionResolver here. I've found it quite helpful when working on a hash of data with partial replacements and default values.

Resources/config/services.yml
```diff
@@ -97,3 +99,19 @@ services:
         class: %ezdemo.installer.demo_installer.class%
         tags:
             - {name: ezplatform.installer, type: demo}
+
+    ezdemo.ez_content_banner_factory:
+        class: %ezdemo.privacy_cookie.banner_factory.class%
+        factory: ["%ezdemo.privacy_cookie.banner_factory.class%", build]
+        arguments:
+            - contentId: %ezdemo.privacy_cookie.content_banner.content_id%
```
bdunogier: Maybe this could be declared as a container variable, and used here. Would be more readable. The array could be in default_settings.yml.

PrivacyCookie/DemoBundleBanner.php
```diff
@@ -0,0 +1,74 @@
...
+            )->path;
+        }
+        else
+        {
+            // fill `Banner` object using defaults
+            $banner->caption = $bannerOptions[ 'bannerCaption' ];
+            $banner->policyPageUrl = $bannerOptions[ 'learnMoreUrl' ];
+        }
+
+        $banner->learnMoreText = $translator->trans( $bannerOptions[ 'learnMoreText' ] );
```
bdunogier: I'd rather get the `learnMoreText` from the Content, and skip the translation part on text defined in configuration.

Resources/config/services.yml
bdunogier: According to the code above, you don't use the `contentId` key, only `pathString`. I would definitely not support both, and we usually use contentId rather than pathString.

PR #152, opened 2015-04-21
EZS-117: Remove duplicate images in Demo bundle for the editor

clash82:
Link to Jira task: https://jira.ez.no/browse/EZS-117


clash82:
This task allows to reduce the size of `demobundle-data` by approximately 1.2 MB.
ping @andrerom @lserwatka @bdunogier

bdunogier:
Can you summarize what this does ?  GH won't show the diff.

clash82:
Images under eZ Mountains page are doubled. This patch changes image paths in some records (9) from `ezcontentobject_attribute` table to rely on the same as paths from image gallery in front-page. Only `ezcontentobject_attribute` table is affected.

crevillo:
uh. but aren't those paths supposed to be generated when uploading the image? is good idea to change that manually?

clash82:
Those files are already there (with generated paths) so I guess it's not a big sin to reuse them :-)

bdunogier:
Err, imho, this creates invalid data. We should try to see what happens when we remove one of the objects, or try to upload a new image. Any effect on the other object using the image ?

Also, what about items in the ezimagefile table ?

clash82:
This task was updated and is more clear now. I'm gonna start all over.

clash82:
Guys, take a look on php/twig - skip sql dump for now
ping @bdunogier @michalpipa @crevillo

Resources/views/full/place.html.twig
```diff
@@ -7,17 +7,23 @@
         </div>
     </div>
     <div class="row">
-        {% if not ez_is_field_empty( content, 'image' ) %}
+        {% if ( relatedImage is empty ) and (ez_is_field_empty( content, 'image' )) %}
```
crevillo: for consistency, could we have spaces before closing and after opening parenthesis all the way?

Controller/PlaceController.php
```diff
@@ -41,6 +41,26 @@ public function listPlaceListAction( Location $location )
...
+    public function showPlaceAction( Location $location, $viewType )
+    {
+        $placeHelper = $this->get( 'ezdemo.place_helper' );
+        $contentService = $this->get( 'ezpublish.api.service.content' );
+
+        return $this->get( 'ez_content' )->viewLocation(
+            $location->id,
+            $viewType,
+            true,
+            [ 'relatedImage' => $placeHelper->getImageFromRelatedObjects( $contentService, $location ) ]
```
crevillo: can't contentService be injected in placeHelper and avoid it pass as parameter from here?

crevillo:
overall, some things i'm thinking
* in the twig templates it see we show relatedImage only if image of the content is empty. But we are getting the related Images always, right? I mean, if we aren't showing those related Images if the content has image, couldn't we add a check in the Action and call the placeHelper only if content has no image?
* Furthermore, and following this check we could have a "image" variable and avoid the if else in the twig templates? 

i mean, something like (more or less...)
```php
<?php
public function showPlaceAction( Location $location, $viewType )
{
         $placeHelper = $this->get( 'ezdemo.place_helper' );
         $content = $this->get( 'ezpublish.api.service.content' )->loadContent( $location->contentId );
         $fieldHelper =  $this->get( 'ezpublish.field_helper' );
         $image = $fieldHelper->isFieldEmpty( $content, 'image' )
            ? $placeHelper->getImageFromRelatedObjects( $contentService, $location )
            : $content;

        return $this->get( 'ez_content' )->viewLocation(
            $location->id,
            $viewType,
            true,
           [ 'imageObject' => $image ]
     );
}
```
and show, in the twig part we could go just for 
```twig
{{ ez_render_field( imageObject, 'image' ) }}
```

Helper/PlaceHelper.php
```diff
@@ -83,6 +85,45 @@ public function getPlaceList( Location $location, $contentTypes, $languages = ar
...
+    public function getImageFromRelatedObjects( ContentService $contentService, Location $location )
+    {
+        $content = $contentService->loadContentByContentInfo( $location->getContentInfo() );
+
+        $relatedObjects = $this->getRelatedObjects(
+            $contentService,
+            $content
+        );
+
+        return reset( $relatedObjects );
```
crevillo: what is that reset for? aren't we still returning an array here?

Resources/views/full/place.html.twig
```diff
@@ -7,17 +7,23 @@
...
-                    )
-                }}
+                {% if not ez_is_field_empty( content, 'image' ) %}
+
+                    {# display assigned place image #}
+                    {{ ez_render_field( content, 'image', { 'attr': { 'class': 'place-image' }, 'parameters': {'alias' : 'imageplacesize'} } ) }}
+                {% else %}
+
+                    {# display related place image #}
+                    {{ ez_render_field( relatedImage, 'image', { 'attr': { 'class': 'place-image' }, 'parameters': {'alias' : 'imageplacesize'} } ) }}
```
crevillo: shouldn't we check if relatedImage is indeed a content? i mean, what if we have no related images? Isn't that the case with demo-clean?

bdunogier: +1. Sounds like an `else if` is needed here.

Helper/PlaceHelper.php
bdunogier: It does not say why it is needed.

Controller/PlaceController.php
bdunogier: Why not. But the helper wasn't changed as part of this PR, was it ?

Helper/PlaceHelper.php
crevillo: i guess is to retuning first element of that array?

Controller/PlaceController.php
crevillo: no, i don't see any changes in placeHelper for this pr. just suggesting/asking if it is a good practice mainly.

Helper/PlaceHelper.php
clash82: > i guess is to retuning first element of that array?

+1

clash82:
Hey @crevillo I like your idea of the `showPlaceAction` modification - very cool, thanks :-)

clash82:
Does anyone have an idea why test fails?

clash82:
Need a strong opinion here. I really have no idea why test fails?!
ping @bdunogier @andrerom

Helper/PlaceHelper.php
```diff
@@ -83,6 +85,45 @@ public function getPlaceList( Location $location, $contentTypes, $languages = ar
     }
 
     /**
+     * Retrieve all objects related to place $content.
+     *
+     * @param \eZ\Publish\Core\Repository\ContentService $contentService
+     * @param \eZ\Publish\API\Repository\Values\Content\Content $content
+     * @return array
+     */
+    public function getRelatedObjects( ContentService $contentService, Content $content )
```
andrerom: private

Helper/PlaceHelper.php
```diff
@@ -83,6 +85,45 @@ public function getPlaceList( Location $location, $contentTypes, $languages = ar
...
+    }
+
+    /**
+     * Returns first image from related objects array.
+     *
+     * @param \eZ\Publish\Core\Repository\ContentService $contentService
+     * @param \eZ\Publish\API\Repository\Values\Content\Location $location
+     * @return \eZ\Publish\API\Repository\Values\Content\Content
+     */
+    public function getImageFromRelatedObjects( ContentService $contentService, Location $location )
```
andrerom: name is misleading (or rather the code is), you don't know if you are actually returning an image here.

Helper/PlaceHelper.php
```diff
@@ -83,6 +85,45 @@ public function getPlaceList( Location $location, $contentTypes, $languages = ar
...
+    {
+        $content = $contentService->loadContentByContentInfo( $location->getContentInfo() );
+
+        $relatedObjects = $this->getRelatedObjects(
```
andrerom: Overkill when you plan to only return one in the end, please assume database (including API calls that implicitly make one) is on another planet in terms of time it takes for an answer.

Helper/PlaceHelper.php
crevillo: good point @andrerom. Actually i think all the code here is supposed to have some specific content in the database. But even this is a demo, user could play with the content and as you say, add a related object to the place not being an image...

Helper/PlaceHelper.php
crevillo: instead of load all the relations the content has, could go for searching in objects related with the content and limit that to 1?

clash82:
Last approach with the smarter solution. Changes made after discussion with @lserwatka.

Please, take a look.
ping @andrerom @bdunogier @crevillo

Resources/views/content_fields.html.twig
```diff
@@ -13,3 +13,13 @@
         {% endif %}
     {% endspaceless %}
 {% endblock %}
+
```
lserwatka: I'm afraid this is wrong. With this template you make global override for all object relation field types. Please remove it and use ez_render_field with proper view.

crevillo:
but arent' those changes in the controller part of this pr anymore? i'm a bit lost now...

lserwatka:
@crevillo controller is not need at all in this task. It is simply change to the Place Content Type where @clash82 replaced Image FieldType with Relation FieldType. It was misunderstanding o the overall concept. This was trivial task.

crevillo:
@lserwatka ok. thanks.

clash82:
OK, final commit added. Once again, please take a look.

ping @andrerom @crevillo @lserwatka @bdunogier

@crevillo images are stored in `demobundle-data` (https://github.com/ezsystems/demobundle-data/pull/1)

andrerom:
@clash82 Please remember to add "[WIP]" in beginning of PR title, and/or use github tasks to signal that something is work in progress so people like @crevillo don't end up reviewing something that is half baked ;)

clash82:
> Please remember to add "[WIP]" in beginning of PR title, and/or use github tasks to signal that something is work in progress so people like @crevillo don't end up reviewing something that is half baked ;)

Thank you for your support. I'll try to remember this :-)

lserwatka:
+1

clash82:
One more `+1` is needed here :-)

ping @andrerom @crevillo

andrerom:
review ping @bdunogier / @lolautruche

michalpipa:
+1

lserwatka:
@andrerom currently there is nothing about caption in the requirements, on the other hand, current solution is tailor made, less DOM structure. I would open new PR if needed and close this one as it does the job mentioned in the user story.

bdunogier:
Looks okay to me.

Is this covered by any BDD test ?

andrerom:
> Is this covered by any BDD test ?


No, no coverage of images*, but then again testing images is hard unless we start to do visual regression testing.
\* https://github.com/ezsystems/DemoBundle/blob/master/Features/Content/place.feature

bdunogier:
It might be worth adding (even afterwards) a test that verifies that we do have an image tag when we expect one. Testing that the image is the right one and stuff is a different story. Based of what I know of the PlatformUI BDD, it should be fairly easy.

clash82:
@bdunogier I will add BDD tests for images later.

clash82:
I've started to write BDD tests for [place] content. I'm gonna push them in separate PR.

Can we merge this one?
ping @andrerom

andrerom:
+1 

> Can we merge this?

Ok, but please from now on start to make sure you take testing into account when estimating, even BDD. Ideally tests should be part of the same PR as otherwise we can't verify that it works before we merge.

Put in a different way, we already aim to have always green branches and test before merge, and aim is to also deal with BDD tests that way as an improvement to our workflow this year. /cc @lserwatka

bdunogier:
Could you please check how this would affect ezsystems/demobundle-data ?

clash82:
> Could you please check how this would affect ezsystems/demobundle-data ?

eZ Mountains images were removed from `demobundle-data`: https://github.com/ezsystems/demobundle-data/pull/1 - local test passed...

PR #153, opened 2015-05-06
[req] Updated components/bootstrap to ~3.3.2

bdunogier:
Fixes the dependency issue Travis was having with bootstrap (https://github.com/ezsystems/ezplatform/pull/15).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "knplabs/knp-menu-bundle": "1.*",
         "zetacomponents/feed": "~1.4",
         "ezsystems/demobundle-data": "~0.1",
-        "components/bootstrap": "3.3.2",
+        "components/bootstrap": "~3.3.2",
         "ezsystems/privacy-cookie-bundle": "~0.1"
     },
     "autoload": {


bdunogier:
@andrerom / @clash82 any objection ?

andrerom:
+1 ( @clash82 context: the hard coded version made composer fail, so we need to open the requirement a little bit here )

clash82:
I would prefer to update it manually from time to time. We've discussed this with @bdunogier. I'll test if everything works and give a feedback.

closed... uh, where's the fire? ;-)

bdunogier:
I'll update it manually then.

andrerom:
> closed... uh, where's the fire? ;-)

sprint demo and release, and this popped up late blocking the whole thing.

Sub/patch release updates are a completely different beast then v2 to v3 (or even v3.2 to v3.3). As per [Semantic Versioning](http://semver.org/), which bootstrap apperently follows now, they are actively trying to avoid introducing any breaks in patch versions. So there might be some micro breaks, but this is demo bundle so we don't care.

PR #154, opened 2015-05-06
Added missing BDD tests for place and place content

clash82:
Please review.

### DIFF
--- a/Features/Content/place.feature
+++ b/Features/Content/place.feature
@@ -3,6 +3,8 @@ Feature: Place
     Scenario: As a visitor, I can see the place page
         Given I am on the "place" page
          Then I see "eZ Mountains" text
+         And I see "Mount Fuji" text
+         And I see a valid thumbnail for image with alternative text "mount fuji"
 
     Scenario: As a visitor, I can see the place list
         Given I am on "/placelist/110"
@@ -11,3 +13,9 @@ Feature: Place
     Scenario: As a visitor, I can see the place list sorted
         Given I am on "/placelist/110/45.6152665/5.2234365/10000"
         Then I see "Mount Ventoux" text
+
+    Scenario: As a visitor, I can see the Mount Fuji page
+        Given I am on "/eZ-Mountains/Mount-Fuji"
+        Then I see "Mount Fuji" text
+        And I see "Mount Fuji, located on Honshu Island, is the highest mountain in Japan at 3,776.24 m (12,389 ft)." text
+        And I see a valid thumbnail for image with alternative text "mount fuji"
--- a/Features/Context/Content/Context.php
+++ b/Features/Context/Content/Context.php
@@ -152,4 +152,32 @@ class Context extends Demo
             );
         }
     }
+
+    /**
+     * Test if image is present on page and is downloadable (`img` tag must contain `alt` attribute)
+     * @Then I (should) see a valid thumbnail for image with alternative text :imageAlternativeText
+     */
+    public function iShouldSeeAVailidThumbnailForImageWithAlternativeText( $imageAlternativeText )
+    {
+        $image = $this->getXpath()->findXpath( "//img[contains(@alt, '" . $imageAlternativeText . "')]" );
+
+        if ( count( $image ) == 0 )
+        {
+            throw new \Exception( sprintf( 'Image with an alternative text `%s` was not found', $imageAlternativeText ) );
+        }
+
+        $file = $image[0]->getAttribute( 'src' );
+
+        $client = new \Guzzle\Http\Client();
+        $request = $client->get( $this->locatePath( $file ) );
+        $response = $request->send();
+
+        Assertion::equalTo(
+            $response->getStatusCode(), 200
+        );
+
+        Assertion::assertRegexp(
+            '/image\/(.*)/', $response->getHeader( 'content-type' )->__toString()
+        );
+    }
 }


joaoinacio:
Looking at the title, shouldn't the PR include only commit 25bcd7f665a50003e4619c14e6240f514d0bd669 ?

clash82:
> Looking at the title, shouldn't the PR include only commit 25bcd7f ?

Yes and no... tests are implemented to check for changes I made in place content, but... they should also work in previous releases.

clash82:
ping @andrerom @bdunogier @michalpipa @lserwatka

Features/Context/Content/Context.php
```diff
@@ -152,4 +152,15 @@ public function breadcrumbHasTheFollowingLinks( TableNode $table )
             );
         }
     }
+
+    /**
+     * Test if image is present on page (`img` tag must contain `alt` attribute)
+     * @Then I (should) see the thumbnail for :imageAlternateText
+     */
+    public function iShouldSeeTheThumbnailFor ( $imageAlternateText )
```
lserwatka: BTW, what CS do we use here? This does not look like PSR-2 to me.

michalpipa: > BTW, what CS do we use here? This does not look like PSR-2 to me.

DemoBundle is still using eZ coding standards.

michalpipa:
:+1:

Features/Content/place.feature
```diff
@@ -11,3 +13,9 @@ Feature: Place
     Scenario: As a visitor, I can see the place list sorted
         Given I am on "/placelist/110/45.6152665/5.2234365/10000"
         Then I see "Mount Ventoux" text
+
+    Scenario: As a visitor, I can see the Mount Fuji page
+        Given I am on "/eZ-Mountains/Mount-Fuji"
+        Then I see "Mount Fuji" text
+        And I see "Mount Fuji, located on Honshu Island, is the highest mountain in Japan at 3,776.24 m (12,389 ft)." text
+        And I see the thumbnail for "mount fuji"
```
bdunogier: I think this should be more verbose: "And I see the thumbnail for image with alternative text "mount fuji".

clash82:
Thumbs up!

ping @andrerom @lserwatka @bdunogier @lolautruche @crevillo @joaoinacio

crevillo:
+1

Features/Content/place.feature
```diff
@@ -11,3 +13,9 @@ Feature: Place
     Scenario: As a visitor, I can see the place list sorted
         Given I am on "/placelist/110/45.6152665/5.2234365/10000"
         Then I see "Mount Ventoux" text
+
+    Scenario: As a visitor, I can see the Mount Fuji page
+        Given I am on "/eZ-Mountains/Mount-Fuji"
+        Then I see "Mount Fuji" text
+        And I see "Mount Fuji, located on Honshu Island, is the highest mountain in Japan at 3,776.24 m (12,389 ft)." text
+        And I see the thumbnail for image with alternative text "mount fuji"
```
bdunogier: We could actually make this a tad more useful. What about:

```
And I see a valid thumbnail for image with alternative text "mount fuji"
```

And then also check in the Context's method that the image path matches what we expect (maybe based on configuration ?)

andrerom: I would say path is implementation, if we can somehow verify image exists, and is not returning 404, that should be enough until we start doing screen shoot regression testing.

bdunogier: What we expect doesn't have to be the actual path. But we could check that it matches whatever is set as the vardir. Or verify that it isn't a 404, yes, and maybe that it is indeed an image.

Features/Content/place.feature
```diff
@@ -11,3 +18,11 @@ Feature: Place
...
+        And Path for thumbnail image with alternative text "mount fuji" should match "/var/ezdemo_site/storage/images/media/images/mount-fuji/552-1-eng-GB/mount-Fuji_imageplacesize.jpg"
+        And File "/var/ezdemo_site/storage/images/media/images/mount-fuji/552-1-eng-GB/mount-Fuji_imageplacesize.jpg" exists
```
andrerom: This (these two lines) is exactly the type of internal info the spec should now know about IMO, it is testing too much (making refactoring and changes to core harder). /cc @bdunogier

andrerom: @clash82 some context:
the argument above is me and @bdunogier argueing over having the path in the test implementation, aka the Context. But having the path in the *Spec*, aka .feature file, is a whole other level. Please take this to heart: *BDD is a Specification by example system, it is not a test framework*, so what you are writing here is a specification for how the system should work, and the feature you are testing is placement, not the internals of the image system.

bdunogier: @andrerom is right, there should be no path here. Sorry if I misled you.

clash82: No problem, but to be sure: can I use hard coded paths in `Context.php` like it is done in `pageIdentifierMap` variable? or should I totally avoid it?

andrerom: It is more ok, but the discuss me and @bdunogier had above was me trying to suggest ways to avoid even that, by having _something_ (you'r more deeply involved in this, so feel free to suggest other forms of verifying image and alternative text*) like 
`
And I see a valid thumbnail for image with alternative text "mount fuji"
`

Where "valid" would imply taking the path and doing a http call to verify it does not return 404.

\* Some input from @ezsystems/qa-team would be good here also

miguelcleverti: the image directory could be set as a parameter in the settings perhaps, this way is not explicit in the BDD and it's not hard coded.

clash82:
Hard coded paths were removed. @andrerom what do you think?

bdunogier:
By the way, about image testing, this is what I wrote a couple days ago:

> What we expect doesn't have to be the actual path. But we could check that it matches whatever is set as the vardir. Or verify that it isn't a 404, yes, and maybe that it is indeed an image.

About the change, it looks good to me. Maybe I'd make the mimetype testing a bit less strict than image/jpeg. What about testing for image/* ?

clash82:
> What about testing for image/* ?

Good point.

clash82:
Test for `image/*` added and passing.
ping @andrerom @bdunogier

bdunogier:
Looks very good, +1 !

clash82:
:heart:

andrerom:
+1 :tada:

clash82:
Thank you :-)

PR #155, opened 2015-05-18
EZP-24346: As a user of Platform and Studio UI, I want to have an almighty content type that includes all default field types of eZ Platform

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-24346

See it [in action](https://www.youtube.com/watch?v=bSlfdjC0PAs).


clash82:
ping @andrerom

andrerom:
$a = @rolandbenedetti : What was the purpose of this task? Should it also contain field types we don't support yet like Tags, deprecated and otherwise missing fiedltypes?

if ( $a == 'it is fine like it is' )
    +1;

clash82:
ping @rolandbenedetti

rolandbenedetti73:
Sorry @andrerom @clash82 I usually see Github notification but this one went through the net.
It should contain only the field type listed in the requirement.
This is a content type that aim at making it simpler to show and explain all the default field types deliver with eZ Platform, it has no real use case purpose beside that one, hence it will only be in the demo bundle to serve that purpose.

PR #156, opened 2015-05-29
EZP-24425 DemoBundle form fields appearing in red

crevillo:
Fixes https://jira.ez.no/browse/EZP-24425

looks like form[field].vars.errors is defined even is there no errors. We need to check if that is empty or not before adding the error class. 

ping @yannickroger @bdunogier @andrerom


andrerom:
+1

crevillo:
ping @clash82 here too.

clash82:
It looks as it should now, good work +1

bdunogier:
+1

PR #157, opened 2015-05-29
EZP-24200: Videos are not played in DemoBundle

crevillo:
Fixes https://jira.ez.no/browse/EZP-24200

but depends on https://github.com/ezsystems/ezpublish-kernel/pull/1273

Now we have a content download controller, this pr changes the template to build the url according to the spec and sets the videoUri to be played.

ping @andrerom @clash82 @bdunogier @yannickroger

### DIFF
--- a/Resources/views/fields/ezbinaryfile_video.html.twig
+++ b/Resources/views/fields/ezbinaryfile_video.html.twig
@@ -3,9 +3,6 @@
 
 {# BinaryFile that contains a video and that we want to display as HTML5 video. #}
 {% block ezbinaryfile_field %}
-    {% set videoUri = '/content/download/' ~ contentInfo.id ~ '/' ~ field.id
-                        ~ '/version/' ~ contentInfo.currentVersionNo ~  "/file/"
-                        ~ field.value.fileName|escape( 'url' ) %}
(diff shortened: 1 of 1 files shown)


andrerom:
+1

clash82:
Works very good, thanks +1

Resources/views/fields/ezbinaryfile_video.html.twig
```diff
@@ -3,9 +3,8 @@
 
 {# BinaryFile that contains a video and that we want to display as HTML5 video. #}
 {% block ezbinaryfile_field %}
-    {% set videoUri = '/content/download/' ~ contentInfo.id ~ '/' ~ field.id
-                        ~ '/version/' ~ contentInfo.currentVersionNo ~  "/file/"
-                        ~ field.value.fileName|escape( 'url' ) %}
+    {% set videoUri = '/content/download/' ~ contentInfo.id ~
```
bdunogier: You should use the route reference for this, like it is done in the default field template.

crevillo:
@bdunogier yes, better as you say. changed as you suggest. ok to have it in one go or do you prefer to define the variable first for readability?

bdunogier:
With our without the variable, it doesn't make much of a difference.

+1

PR #158, opened 2015-06-01
Implemented EZP-24435: Use a dedicated template to theming feedback forms

crevillo:
Implements: https://jira.ez.no/browse/EZP-24435

The goal of this PR is to have a dedicate template for theming feedback forms. Thanks to Symfony form theming capabilities, we can add a dedicate template for every form. 

Further more, as all the fields in the forms are showed in the same way, we can avoid the html we actually have inside the form and let the theming template do its jobs. 

This way, if for every reason we need to change the class of the labels for these forms we will only need to do one change and no one per field. 

Feedback welcome

ping @andrerom @yannickroger @clash82 @bdunogier

### DIFF
--- a/Resources/views/forms/feedback_form.html.twig
+++ b/Resources/views/forms/feedback_form.html.twig
@@ -1,0 +1,22 @@
+{% block form_row -%}
+    {% set class = form.vars.label_attr.class is defined ?  form.vars.label_attr.class : 'control-label' %}
+
+    <div class="control-group{% if not form.vars.valid %} error{% endif %}">
+        {{- form_label(form) -}}
+        <div class="controls">
+            {{- form_widget(form) -}}
+            {{- form_errors(form) -}}
+        </div>
+    </div>
+{%- endblock form_row %}
+
+{% block form_errors %}
+    {% if errors|length > 0 %}
+        <span class="help-inline">
+            {% for error in errors %}
+                {{ error.message }}
+                {% if not loop.last %},{% endif %}
+            {% endfor %}
+        </span>
+    {% endif %}
+{% endblock form_errors %}
--- a/Resources/views/full/feedback_form.html.twig
+++ b/Resources/views/full/feedback_form.html.twig
@@ -1,21 +1,6 @@
 {% extends "eZDemoBundle::pagelayout.html.twig" %}
 
-{# Customizing the error message to add a css class since form_errors doesn't accept parameters #}
-{% form_theme form _self %}
-
-{% block form_errors %}
-    {% if errors|length > 0 %}
-        <span class="help-inline">
-            {% for error in errors %}
-                {{ error.message }}
-                {% if not loop.last %},{% endif %}
-            {% endfor %}
-        </span>
-    {% endif %}
-{% endblock form_errors %}
-
 {% block content %}
-
 <section class="content-view-full">
     <div class="row">
         <div class="col-md-8">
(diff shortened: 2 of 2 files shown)


clash82:
It's more clear now +1

andrerom:
+1, review ping @lolautruche / @glye

Resources/views/forms/feedback_form.html.twig
```diff
@@ -0,0 +1,22 @@
+{% block form_row -%}
+    {% set class = form.vars.label_attr.class is defined ?  form.vars.label_attr.class : 'control-label' %}
```
lolautruche: Extra WS after `?`

Resources/views/full/feedback_form.html.twig
```diff
@@ -34,61 +19,42 @@
...
+                    'label': 'Message',
+                    'attr': { 'class': 'col-md-8', 'rows': 10 }
+                } ) }}
+
+                {{ form_row( form.save, {
+                    'label': "Send form",
+                    'attr': { 'class': 'btn btn-warning pull-right' }
+                } ) }}
 
                 {{ form_end( form ) }}
```
lolautruche: Watch out here. As you redefine form_row, hidden fields will probably also be affected (e.g. csrf token field)

Resources/views/forms/feedback_form.html.twig
crevillo: Fixed, thanks

Resources/views/full/feedback_form.html.twig
crevillo: @lolautruche  csrf token is there. ```form_end( form )``` takes care of adding. it will also add all the fields not showed before. Ref http://symfony.com/doc/current/book/forms.html
```
form_end(form)
Renders the end tag of the form and any fields that have not yet been rendered, in case you rendered each field yourself. This is useful for rendering hidden fields and taking advantage of the automatic CSRF Protection.
```

Anyway what is done here is just use ```form_row``` for the fields instead of ```form_label```, ```form_widget```, ```form_errors```.

lolautruche: I'm not saying csrf token field won't be here. I'm just saying that its rendering will also use the markup defined in your theme, as `form_rest` block, which is called by `form_end` uses `form_row` as well. See [`form_div_layout.html.twig`](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L299-L305).
This is probably not something we want...

crevillo: understood, i see your point. Anyway the html I get for csrf_token is ```<input type="hidden" id="ezdemo_feedback_ezxform_token" name="ezdemo_feedback[ezxform_token]" value="Qfis-4hg1UIDPGGjiidOu2X92IBxs272tq95qOmLZ5I" />``` it seems my form_row is not applied there. Trying to know why.

crevillo: ok, reason is https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L253. 
there is a dedicate block for hidden_row, that's why csfr_token doesn't try to print label . 

Anyway i re-added 
```twig
  <div class="row">
       {{ form_errors( form ) }}
  </div>
```
to show possible errors coming from it.

Resources/views/forms/feedback_form.html.twig
```diff
@@ -0,0 +1,22 @@
...
+{% block form_errors %}
+    {% if errors|length > 0 %}
+        <span class="help-inline">
+            {% for error in errors %}
+                {{ error.message }}
+                {% if not loop.last %},{% endif %}
+            {% endfor %}
+        </span>
+    {% endif %}
+{% endblock form_errors %}
```
lolautruche: Missing empty line

lolautruche:
+1

Resources/views/forms/feedback_form.html.twig
crevillo: added, thanks.

Resources/views/full/feedback_form.html.twig
```diff
@@ -34,61 +20,46 @@
...
-                    {{ form_label( form.message, 'Message'|trans,  { 'attr': {'class': 'control-label'}} ) }}
-                    <div class="controls">
-                        {{ form_widget( form.message, { 'attr': {'class': 'col-md-8', 'rows': '10'} } ) }}
-                        {{ form_errors( form.message ) }}
-                    </div>
-                </div>
-
-                {{ form_widget( form.save, { 'label': "Send form"|trans, 'attr': { 'class': 'btn btn-warning pull-right' } } ) }}
+                {{ form_row( form.firstName, {
+                    'label': 'First Name',
```
yannickroger: @crevillo : The `|trans` is not needed anymore ?

crevillo: @yannickroger: default form_label from symfony label will take care of it 
https://github.com/symfony/symfony/blob/2.7/src/Symfony/Bridge/Twig/Resources/views/Form/form_div_layout.html.twig#L223

yannickroger: Great :)

crevillo:
gentle ping.

yannickroger:
+1

yannickroger:
Thanks @crevillo

yannickroger:
Can you just add the issue number in the commit message and ping me, I'll merge it asap

crevillo:
done @yannickroger

yannickroger:
@crevillo the commit message ?

crevillo:
sorry @yannickroger , i missunderstood. i've changed the commit and rebased into master.

PR #160, opened 2015-06-04
Fix: EZP-24477 Some main menu items dissapear when you're logged

crevillo:
Fixes https://jira.ez.no/browse/EZP-24477, https://jira.ez.no/browse/EZP-24413

By default ```$query->limit``` is 10, so if you don't raise that, the query only returns the 10 first elements it finds in first and second level. 
That's because two last main menu items are missing.

In this patch, I raise the limit to 25. But how could we build a query without limit? i mean, a query that will return all the elements without the need to set $query->limit?

ping @andrerom @bdunogier @clash82

### DIFF
--- a/Menu/Builder.php
+++ b/Menu/Builder.php
@@ -142,6 +142,7 @@ class Builder
             )
         );
         $query->sortClauses = array( new Query\SortClause\Location\Path() );
+        $query->limit = 25;
         $query->performCount = false;
 
         return $this->searchService->findLocations( $query )->searchHits;


wizhippo:
@crevillo you can't since the search api was updated.  I believe this was to force a limit as some search providers could have issues without one.   You could page through the results though :(

crevillo:
@wizhippo what you mean by page throught the results?

wizhippo:
https://jira.ez.no/browse/EZP-24413 is also affected by the search limit.

crevillo:
ah, sorry, didn't see that one...

wizhippo:
@crevillo the query has a limit and offest.  You would have to get the total count then query at offset + limit until you hit the end.

crevillo:
wizhippo but that will imply more queries that you don't really need, right?

wizhippo:
I agree.  But this was changed in https://github.com/ezsystems/ezpublish-kernel/commit/8e6e38e1589ee75085dba38e9b5cc450d4224fb6.  Paging is forced by design now.

crevillo:
i see. yes, my patch is only a workaround for this. it doesn't assure you will get the entire menu if the menu grows. Let's wait the others to see what they think.

andrerom:
ping @pspanja BC break mister :/

crevillo:
gentle ping @pspanja

andrerom:
Closing the default limit has been raised to 25 :)

PR #161, opened 2015-06-04
Fix EZP-24478: Partner page can't be seen without legacy enabled

crevillo:
Fix https://jira.ez.no/browse/EZP-24478

Make possible to access and view Partner page without legacy stack enabled

It just ass a line view for files for it. 

Added also a feature for this page. 

Let me know if it's ok

ping @andrerom @miguelcleverti @joaoinacio @bdunogier

### DIFF
--- a/Features/Context/Demo.php
+++ b/Features/Context/Demo.php
@@ -33,7 +33,8 @@ class Demo extends Context
             "place" => "/eZ-Mountains",
             "services" => "/Shopping/Services",
             "products" => "/Shopping/Products",
-            "resources" => "/Getting-Started/Resources"
+            "resources" => "/Getting-Started/Resources",
+            "partner" => "/Partner",
         );
 
         $this->mainAttributes += array(
--- a/Features/Content/partner.feature
+++ b/Features/Content/partner.feature
@@ -1,0 +1,23 @@
+Feature: Partner
+    Scenario: A visitor can't see "Partner" in the main menu
+        Given I am not logged in
+         Then on "main navigation" I shouldn't see links:
+             | Partner |
+
+    Scenario: A visitor can't access "Partner" page
+        Given I am not logged in
+        Given I am on the homepage
+         When I go to "partner" page
+         Then I see "Login" text
+
+    Scenario: As administrator, I can see "Partner" in the main menu
+        Given I am logged as an "administrator"
+        Given I am on the "home" page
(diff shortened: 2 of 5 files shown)


andrerom:
+1

yannickroger:
+1
Fixing a bug by implementing a new feature and include BDD, it is not OK, it is GREAT :)
ping @bdunogier

crevillo:
gentle ping @bdunogier

Features/Context/Demo.php
```diff
@@ -33,7 +33,8 @@ public function __construct()
             "place" => "/eZ-Mountains",
             "services" => "/Shopping/Services",
             "products" => "/Shopping/Products",
-            "resources" => "/Getting-Started/Resources"
+            "resources" => "/Getting-Started/Resources",
+            "partner" => "/Partner"
```
bdunogier: Please add a comma after the last item, to limit the diff for the next change.

bdunogier:
This partially implements https://jira.ez.no/browse/EZP-24466, as a bonus (integrate native binaryfile/media download into the demobundle)

+1, good job on the BDD :-)

crevillo:
thanks @bdunogier . yeah, i noticed we don't have view for "file" in demobundle. was on my todo task. added the commas, here too.  https://github.com/ezsystems/DemoBundle/pull/161/files#diff-9cccdae977ea8fdb469bf728d7a4f264R40

yannickroger:
Thanks @crevillo

bdunogier:
By the way, does the look and feel of the `line` view for `file` Content Type match what we had in legacy ezdemo ?

crevillo:
@bdunogier that's a very good question :). I'll try to explain. i was doing having another ezpublish-community with legacy enabled installation. 
There, there was the ezdemo extension having [this file](https://github.com/ezsystems/ezdemo/blob/master/packages/ezdemo_extension/ezextension/ezdemo/design/ezdemo/override/templates/line/file.tpl)

But for whatever reason, with legacy enabled in that installation, this template was not loaded at all. Instead [this other one](https://github.com/ezsystems/ezpublish-legacy/blob/master/design/base/override/templates/line/file.tpl). And this was the one i took as base for the twig one. 
But let me know if the other one should be the based instead.

bdunogier:
Well, it's a matter of look and feel. If the result is okay, then there is no need to change it.

PR #162, opened 2015-06-05
Fix EZP-24482 - eZ Demo top menu doesn't appear until added a new folder after EZP-22361

pedroresende:
After updating DemoBundle to use the latest Bootstrap https://jira.ez.no/browse/EZP-22361 a regression has been introduced https://jira.ez.no/browse/EZP-20108

https://jira.ez.no/browse/EZP-24482

### DIFF
--- a/Resources/public/css/bootstrap.css
+++ b/Resources/public/css/bootstrap.css
@@ -8994,7 +8994,7 @@ del:hover {
   border: none;
 }
 .navbar-header {
-  display: none;
+  min-height: 33px;
 }
 .navbar-header,
 .navbar-top {
--- a/Resources/public/less/navbar.less
+++ b/Resources/public/less/navbar.less
@@ -113,7 +113,7 @@
 }
 
 .navbar-header {
-  display: none;
+  min-height: 33px;
 }
 
 .navbar-header, .navbar-top {


yannickroger:
You should update the description with links to issues instead of just names.

pedroresende:
@yannickroger updated, thanks :)

yannickroger:
ping @clash82 ;) our bootstrap hero.

clash82:
superman

Tested and approved +1

SylvainGuittard:
The main issue is that we are using a customized bootstrap.css. We will face the same issue in the future if we will upgrade bootstrap.

Why not using the standard bootstrap.css and creating a new file (ezdemo.css) for our custom code?

crevillo:
+100 to @SylvainGuittard suggestion.

clash82:
> The main issue is that we are using a customized bootstrap.css. We will face the same issue in the future if we will upgrade bootstrap.

`bootstrap.css` is a compiled version of origin bootstrap less components and our customized version (take a look on `bootstrap.less` file to see how it works). This was part of my [super-hero bootstrap task](https://jira.ez.no/browse/EZP-22361).

I agree that it's not a perfect solution but it cover most of our current needs.

pedroresende:
@clash82 @yannickroger ping....

clash82:
+2

yannickroger:
+1 final ping @SylvainGuittard

SylvainGuittard:
+1

PR #164, opened 2015-06-09
EZP-24466: Add views for file and video content types

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-24466

### DIFF
--- a/Resources/views/full/file.html.twig
+++ b/Resources/views/full/file.html.twig
@@ -1,0 +1,30 @@
+{% extends "eZDemoBundle::pagelayout.html.twig" %}
+
+{% block content %}
+    <section class="content-view-full">
+        <article class="class-file">
+
+            <div class="attribute-header">
+                <h1>{{ ez_content_name( content ) }}</h1>
+            </div>
+
+            {% if not ez_is_field_empty( content, 'description' ) %}
+                <div class="attribute-long">
+                    {{ ez_render_field( content, 'description' ) }}
+                </div>
+            {% endif %}
+
+            <div class="attribute-file">
+                {{ ez_render_field( content, 'file', { attr: { 'class': 'btn btn-warning' } } ) }}
+            </div>
+
+            <div class="attribute-tags">
+                {{ ez_render_field( content, 'tags', { 'attr': { 'class' : 'tags-wrapper' } } ) }}
+            </div>
+
+            <div class="attribute-star-rating">
+                {{ render( controller( 'eZDemoBundle:Demo:viewStarRating', { 'location': location } ) ) }}
+            </div>
+        </article>
+    </section>
+{% endblock %}
--- a/Resources/views/full/video.html.twig
+++ b/Resources/views/full/video.html.twig
@@ -1,0 +1,38 @@
+{% extends "eZDemoBundle::pagelayout.html.twig" %}
+
+{% block content %}
+    {% set videoPath = path( ez_route( 'ez_content_download', { 'content': content, 'fieldIdentifier': 'file' } ) ) %}
+    {% set videoType = ez_field_value( content, 'file' ).mimeType|default( 'video/mp4' ) %}
+
+    <script>
+        _V_.options.flash.swf = '{{ asset( 'bundles/ezdemo/flash/video-js.swf' ) }}';
+    </script>
+
+    <section class="content-view-full">
+        <article class="class-video">
+            <div class="attribute-header">
+                <h1>{{ ez_content_name( content ) }}</h1>
+            </div>
+
+            <div class="attribute-video">
+                {{ ez_render_field( content, 'file', {
+                    'attr': { 'id': 'video_' ~ content.id, 'class': 'video-js vjs-default-skin', 'data-setup': '{}' },
+                    'parameters': { 'width': '770', 'height': '318', 'videoType': videoType },
+                    'template': 'eZDemoBundle:fields:ezbinaryfile_video.html.twig'
+                } ) }}
+            </div>
+
+            <div class="attribute-download">
+                <p><a class="btn btn-mini btn-warning pull-right" href="{{ videoPath }}">{{ 'Download'|trans }}</a></p>
+            </div>
+
+            <div class="attribute-long">
+                {{ ez_render_field( content, 'caption' ) }}
+            </div>
+
+            <div class="attribute-star-rating">
+                {{ render( controller( 'eZDemoBundle:Demo:viewStarRating', { 'location': location } ) ) }}
+            </div>
+        </article>
+    </section>
+{% endblock %}
--- a/Resources/views/line/video.html.twig
+++ b/Resources/views/line/video.html.twig
@@ -1,0 +1,27 @@
+{% set videoType = ez_field_value( content, 'file' ).mimeType|default( 'video/mp4' ) %}
+
+<div class="content-view-line">
+    <article class="class-video row">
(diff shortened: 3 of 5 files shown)


Resources/views/line/video.html.twig
```diff
@@ -0,0 +1,27 @@
+{% set videoType = content.getField( 'file' ).value.mimeType|default( 'video/mp4' ) %}
+
+<div class="content-view-line">
+    <article class="class-video row">
+        <div class="col-md-4">
+            <div class="attribute-video">
+                {{ ez_render_field( content, 'file', {
+                    'attr': { 'id': 'video_' ~ content.id, 'class': 'video-js vjs-default-skin', 'data-setup': '{}' },
+                    'parameters': { 'width': '100%', 'height': '113', 'videoType': videoType },
```
crevillo: do we want to show the videos in line view? Supposing a list with 10 videos this will add 10 videos to the page...

Resources/views/line/video.html.twig
```diff
@@ -0,0 +1,27 @@
+{% set videoType = content.getField( 'file' ).value.mimeType|default( 'video/mp4' ) %}
+
```
crevillo: Also, if we add https://github.com/ezsystems/DemoBundle/pull/164/files#diff-7b44bced1c149a4c107e0bb40e7e7e5bR9 to full video, wouldn't we need something similar for line if we plan show videos in line view?

Resources/views/full/file.html.twig
```diff
@@ -0,0 +1,34 @@
...
+            <div class="attribute-file">
+                <p><a href="{{ downloadPath }}" class="btn btn-warning">{{ fileName }} {{ fileSize }}</a></p>
+            </div>
+
+            <div class="attribute-tags">
+                {{ ez_render_field( content, 'tags', { 'attr': { 'class' : 'tags-wrapper' } } ) }}
+            </div>
+
+            <div class="attribute-star-rating">
+                {{ render( controller( 'eZDemoBundle:Demo:viewStarRating', { 'location': location } ) ) }}
```
crevillo: where is this viewStarRating action defined?

Resources/views/full/file.html.twig
```diff
@@ -0,0 +1,34 @@
+{% extends "eZDemoBundle::pagelayout.html.twig" %}
+
+{% block content %}
+    {% set fileSize = content.getField( 'file' ).value.fileSize|ez_file_size( 4 ) %}
```
blankse: You should use the ez_field_value Twig helper to get the current language.
Otherwise you get always the main language.

Resources/views/full/video.html.twig
```diff
@@ -0,0 +1,38 @@
...
+        <article class="class-video">
+            <div class="attribute-header">
+                <h1>{{ ez_content_name( content ) }}</h1>
+            </div>
+
+            <div class="attribute-video">
+                {{ ez_render_field( content, 'file', {
+                    'attr': { 'id': 'video_' ~ content.id, 'class': 'video-js vjs-default-skin', 'data-setup': '{}' },
+                    'parameters': { 'width': '770', 'height': '318', 'videoType': videoType },
+                    'template': 'eZDemoBundle:fields:ezbinaryfile_video.html.twig'
```
crevillo: This is probably offtopic, but why are we using binary field to store a video in a content of type video? Why don't go for a media field instead?

Resources/views/full/file.html.twig
clash82: Yep, you're right. Thanks.

Resources/views/full/file.html.twig
clash82: It was already used by Articles.

crevillo: yes, but what it does? i can't see a viewStarRating action at https://github.com/ezsystems/DemoBundle/blob/master/Controller/DemoController.php

clash82: https://github.com/ezsystems/DemoBundle/blob/master/Controller/DemoController.php#L235 ?

Resources/views/full/video.html.twig
clash82: It was already here. @lolautruche can give you more answer for that.

crevillo: oops, sorry, it it's there i said nothing :).

Resources/views/line/video.html.twig
clash82: ok, but what's wrong with that? legacy templates was working exactly the same way :)

crevillo: well, i guess is more work for the browser and more load time to the whole thing. but i'm ok with this.

Resources/views/line/video.html.twig
clash82: Content may contain lot of video files, we should not declare this settings every time. It should be added only once on every page (btw, this setting was missing in legacy too).

clash82:
ping @SylvainGuittard @wizhippo @yannickroger

Resources/views/line/video.html.twig
crevillo: yes, but my point is if you're in a folder you are not adding that. you only list the videos but never set that setting you set in full view, right? 
in legacy we had something like ```run-once``` but dunno if we have something similar here.

clash82: We don't have such extension in Twig. The only way to bypass this issue is to add script on every template.

clash82:
ping ping ping :)

Resources/views/full/file.html.twig
```diff
@@ -0,0 +1,34 @@
...
+            </div>
+
+            {% if not ez_is_field_empty( content, 'description' ) %}
+                <div class="attribute-long">
+                    {{ ez_render_field( content, 'description' ) }}
+                </div>
+            {% endif %}
+
+            <div class="attribute-file">
+                <p><a href="{{ downloadPath }}" class="btn btn-warning">{{ fileName }} {{ fileSize }}</a></p>
```
bdunogier: Why not use `ez_render_field` here ?

clash82: Yep, it's a good idea, thank you.

kamilmusial:
:+1:

michalpipa:
:+1:

clash82:
ping @andrerom

Resources/views/line/video.html.twig
SylvainGuittard: @crevillo I agree. It's a lot of work for the browser if the page has 10 videos. But it's not the case for the DemoBundle. There is only one video.

SylvainGuittard:
+1

Resources/views/line/video.html.twig
crevillo: @SylvainGuittard  well, anyway, we still miss the script where we set the ```_V_.options.flash.swf```  like it's done in full view, don't we?

andrerom:
awaiting travis.

clash82:
It's a good day for merge. Travis is not yelling...

clash82:
ping @andrerom :-)

andrerom:
you should ping others I think, I don't always have time to stay instant with github notifications :)

clash82:
> you should ping others I think, I don't always have time to stay instant with github notifications :)

...it is always good to follow the Boss and ask him for an advice… when you feel lost ;)

Thank you.

PR #167, opened 2015-07-02
Disable demo-clean option

andrerom:
ping @bdunogier 

heads up @lserwatka

demo-clean is missing ezcontentobject_name data, and maybe other things as well. This causes exception after recent kernel Legacy SE change to content loading sql to reduce high amounts of rows -> low perf -> high mem-use -> reaching mem limits.


bdunogier:
See #168 for an alternative.

andrerom:
Closing in favor of #168

PR #170, opened 2015-07-30
Fix EZP-24677: Use sorting order from location

wizhippo:
Perhaps this should be moved into a sortclause helper?

Better solution would be to expose LocationService::getSortClauseBySortField as public or find a better home for it.

### DIFF
--- a/Helper/CriteriaHelper.php
+++ b/Helper/CriteriaHelper.php
@@ -9,7 +9,9 @@
 namespace EzSystems\DemoBundle\Helper;
 
 use eZ\Publish\API\Repository\Values\Content\Location;
+use eZ\Publish\API\Repository\Values\Content\Query;
 use eZ\Publish\API\Repository\Values\Content\Query\Criterion;
+use eZ\Publish\API\Repository\Values\Content\Query\SortClause;
 use eZ\Publish\Core\Base\Exceptions\InvalidArgumentType;
 use DateTime;
 use DateInterval;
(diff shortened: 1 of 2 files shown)


Helper/CriteriaHelper.php
```diff
@@ -149,4 +151,48 @@ public function generateListBlogPostCriterion( Location $location, array $viewPa
         return new Criterion\LogicalAnd( $criteria );
     }
 
+    public function getSortClauseBySortField( $sortField, $sortOrder = Location::SORT_ORDER_ASC )
```
andrerom: We should rather expose this from LocationService then copy it yet another time *(exists in LocationService, UserService and possible other places)*. What is needed to be able to expose this API on LocationService interface cleanly is to be able to type hint Criteria and SortClauses that are supported by Location Search.

andrerom:
Besides the code duplication issue, one of the behat test suites also seems to fail for some reason.

clash82:
Failing behat tests due to depricated criterion clause (https://jira.ez.no/browse/EZP-24730). Fixed now and should pass after rebase. Besides this ezrobot switched to PSR-2 :)

Helper/CriteriaHelper.php
wizhippo: @andrerom Besides making getSortClauseBySortField public in LocationService, can you help me understand what you are suggesting for the type hinting? Also why couldn't UserService and the others also use getSortClauseBySortField if exposed as public instead of repeating ourselves?

clash82:
Hi @wizhippo, something's not right with Travis. Can you push it again?

andrerom:
Regarding `getSortClauseBySortField`: this should be on LocationService in kernel, however a slight blocker is that it is currently not possible to type hint Location vs Content Criteria and SortClauses, which is also needed for proper validation. I would suggest we discuss this with @pspanja on Slack next week to find a solution.

PR #172, opened 2015-08-19
EZP-24717: Image alt attribute should not be shown when empty

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-24717

### DIFF
--- a/Resources/views/fields/ezimage_simple.html.twig
+++ b/Resources/views/fields/ezimage_simple.html.twig
@@ -4,10 +4,11 @@
  #}
 {% block ezimage_field %}
     {% if not ez_is_field_empty( content, field ) %}
+        {% set image_alt = parameters.alt|default( field.value.alternativeText ) %}
         {% set image_alias = ez_image_alias( field, versionInfo, parameters.alias|default( 'original' ) ) %}
         {# if image alias is not found, null is returned, an error is added in logs and the image is not displayed #}
         {% if image_alias %}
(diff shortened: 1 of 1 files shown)


clash82:
@ezrobot you're so mysterious today...

andrerom:
+1, as for ezrobot it is probably that cs needs to be fixed on this repo.

clash82:
Thanks @andrerom 

ping @kamilmusial @SylvainGuittard

crevillo:
+0.5. Alt attribute was required in html and there are some discussions and people arguing why html5 has made this optional... 
http://www.paciellogroup.com/blog/2008/02/alt-in-html5-required-to-be-or-not-to-be/
http://adrianroselli.com/2011/04/image-alt-attributes-not-always.html
Also, as said in http://stackoverflow.com/questions/7729918/is-value-for-alt-atrribute-not-required-in-html5-to-pass-validation, it not specified a screenreader will read the full path of the image instead of say nothing...

kamilmusial:
+1

clash82:
@crevillo thanks for details. DemoBundle is built on top of HTML5 so as you mention, `alt` attribute is optional.

crevillo:
yes, but still... +0.5 :)

clash82:
:beer: 0,5 - like them too ;)

SylvainGuittard:
+1

clash82:
ping @andrerom :)

PR #174, opened 2015-08-24
EZP-24566: Add "Rich Text" field to Almighty content type in demo bundle

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-24566

### DIFF
--- a/Resources/installer/sql/demo_data.sql
+++ b/Resources/installer/sql/demo_data.sql
@@ -370,7 +370,7 @@ INSERT INTO `ezcontentclass` VALUES (0,'<title>',1423236421,14,36,'forums',2,1,3
 INSERT INTO `ezcontentclass` VALUES (0,'<name>',1423236421,14,37,'video',2,0,3,1440579755,14,'b38417e8194fb8f893ca918d297b4fa8','a:2:{s:6:\"eng-GB\";s:121:\"Defines a structure for storing video files. The Video content type is used to hold a file and a description of the file.\";s:16:\"always-available\";s:6:\"eng-GB\";}','a:2:{s:6:\"eng-GB\";s:5:\"Video\";s:16:\"always-available\";s:6:\"eng-GB\";}',1,1,'',0);
 INSERT INTO `ezcontentclass` VALUES (0,'<name>',1423236421,14,38,'place',2,0,3,1440579567,14,'861f02510226434178f3859242fe40d7','a:2:{s:6:\"eng-GB\";s:80:\"Defines a structure for storing a Place object (includes Geolocation attribute).\";s:16:\"always-available\";s:6:\"eng-GB\";}','a:2:{s:6:\"eng-GB\";s:5:\"Place\";s:16:\"always-available\";s:6:\"eng-GB\";}',1,1,'',0);
 INSERT INTO `ezcontentclass` VALUES (0,'<name>',1423236421,14,39,'place_list',2,1,3,1440404785,14,'e3461a7a3d2b6c0d291be6becfb6127e','a:2:{s:6:\"eng-GB\";s:60:\"Here you set up a structure for storing Place objects in it.\";s:16:\"always-available\";s:6:\"eng-GB\";}','a:2:{s:6:\"eng-GB\";s:10:\"Place list\";s:16:\"always-available\";s:6:\"eng-GB\";}',1,1,'',0);
-INSERT INTO `ezcontentclass` VALUES (0,'<text_line>',1431942809,14,40,'almighty',2,0,3,1440580164,14,'b04ae3a76963fe8389b7f3de863f4c74','a:2:{s:6:\"eng-GB\";s:38:\"All default field types of eZ Platform\";s:16:\"always-available\";s:6:\"eng-GB\";}','a:2:{s:6:\"eng-GB\";s:8:\"Almighty\";s:16:\"always-available\";s:6:\"eng-GB\";}',1,1,'',0);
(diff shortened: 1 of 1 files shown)


andrerom:
Difficult rebase on this one? ping me on when you think this can be done considering your agenda with @SylvainGuittard. Just need to know as I plan to add a tag after this is in.

clash82:
@andrerom sorry for that. I had some problems with my eZ installation :) This task was not so easy due to `ezrichtext` is not supported by legacy and ContentType editing is not yet implemented in PlatformUI.

clash82:
ping @andrerom

Resources/installer/sql/demo_data.sql
```diff
@@ -608,7 +609,6 @@ INSERT INTO `ezcontentclass_classgroup` VALUES (37,0,3,'Media');
 INSERT INTO `ezcontentclass_classgroup` VALUES (38,0,1,'Content');
 INSERT INTO `ezcontentclass_classgroup` VALUES (39,0,1,'Content');
 INSERT INTO `ezcontentclass_classgroup` VALUES (40,0,1,'Content');
-INSERT INTO `ezcontentclass_classgroup` VALUES (40,1,1,'Content');
```
andrerom: should not be here I think

Resources/installer/sql/demo_data.sql
```diff
@@ -649,7 +649,6 @@ INSERT INTO `ezcontentclass_name` VALUES (37,0,3,'eng-GB','Video');
 INSERT INTO `ezcontentclass_name` VALUES (38,0,3,'eng-GB','Place');
 INSERT INTO `ezcontentclass_name` VALUES (39,0,3,'eng-GB','Place list');
 INSERT INTO `ezcontentclass_name` VALUES (40,0,3,'eng-GB','Almighty');
-INSERT INTO `ezcontentclass_name` VALUES (40,1,3,'eng-GB','Almighty');
```
andrerom: should not be here I think

Resources/installer/sql/demo_data.sql
```diff
@@ -1399,6 +1398,7 @@ INSERT INTO `ezcontentobject_attribute` VALUES (0,332,126,0,NULL,'<?xml version=
 INSERT INTO `ezcontentobject_attribute` VALUES (0,333,126,0,NULL,'','ezcountry',787,'eng-GB',2,0,'',1);
 INSERT INTO `ezcontentobject_attribute` VALUES (0,334,126,0,NULL,'','ezkeyword',788,'eng-GB',2,0,'',1);
 INSERT INTO `ezcontentobject_attribute` VALUES (0,335,126,0,0,'','ezgmaplocation',789,'eng-GB',2,0,'',1);
+INSERT INTO `ezcontentobject_attribute` VALUES (0,336,126,NULL,NULL,'<?xml version=\"1.0\" encoding=\"utf-8\"?>\n<section xmlns:image=\"http://ez.no/namespaces/ezpublish3/image/\" xmlns:xhtml=\"http://ez.no/namespaces/ezpublish3/xhtml/\" xmlns:custom=\"http://ez.no/namespaces/ezpublish3/custom/\"/>\n','ezxmltext',790,'eng-GB',2,0,'',1);
```
andrerom: is the content here valid? ping @pspanja

Resources/installer/sql/demo_data.sql
clash82: From what I understand is that second `insert` value here:

```sql
INSERT INTO `ezcontentclass_classgroup` VALUES (40,1,1,'Content');
```

is `contentclass_version` and since eZ Publish decided to remove unused version (draft?) it's ok.

Resources/installer/sql/demo_data.sql
clash82: Same as https://github.com/ezsystems/DemoBundle/pull/174#discussion_r38073037

Resources/installer/sql/demo_data.sql
andrerom: You are right, this one is ok.

Resources/installer/sql/demo_data.sql
andrerom: ok, clear :)

Resources/installer/sql/demo_data.sql
pspanja: It looks correct for `ezxmltext`, but I guess it should be for `ezrichtext` instead.

clash82: Thanks, I'll change it and see if works.

andrerom: @pspanja but the content should be different as well right? (not `//ez.no/namespaces/ezpublish3/custom/` as we use docbook in `ezrichtext`)

pspanja: Correct, empty value in `ezrichtext` internal format can be found here: https://github.com/ezsystems/ezpublish-kernel/blob/master/eZ/Publish/Core/FieldType/RichText/Value.php#L21

andrerom: @clash82 how did you change the db? manual or via api? maybe we have a bug somewhere.

clash82: I've changed ContentType manually by adding `XmlBlock` and changing `ezxmlblock` to `ezrichtext` in `contentclass_attributes` table `data_type_string` column.

clash82: `data_type_string` and `data_text` changed. I can confirm that it works.

andrerom:
+1

bdunogier:
+1

PR #176, opened 2015-08-27
Switch coding standards to PSR-2

clash82:


clash82:
ping @michalpipa @lolautruche

michalpipa:
There were some problems, but finally it's done.

+1

clash82:
Very good, indeed :+1: :)

andrerom:
+1

PR #177, opened 2015-08-31
EZP-24745: Make SolrBundle ready for release

andrerom:
- [x] Config change for solr, and remove stash config as it is moved to config.yml
- [x] Fix demo data issue with invalid url on a file, turned out to be media file from almighty content recently added.


andrerom:
Ready for review @bdunogier @glye @yannickroger

glye:
+1

PR #179, opened 2015-09-03
EZP-24762: Remove unused "New Almighty" content object from demo_data.sql

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-24762

### DIFF
--- a/Resources/installer/sql/demo_data.sql
+++ b/Resources/installer/sql/demo_data.sql
@@ -174,7 +174,6 @@ INSERT INTO `ezcobj_state_link` VALUES (118,1);
 INSERT INTO `ezcobj_state_link` VALUES (119,1);
 INSERT INTO `ezcobj_state_link` VALUES (120,1);
 INSERT INTO `ezcobj_state_link` VALUES (121,1);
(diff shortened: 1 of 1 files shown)


andrerom:
Missing change to ezmedia table, you can cherry pick my commit 782c7dc5b9fe9fc4e80d92e2105962d500cfe4aa.

There might also be data in other tables, I'll check the original PR.

andrerom:
It's close to impossible to see the diff of the original PR given had no new lines. Can you try to have a look what was added somehow? to double check that we are not leaving any cruft in the database.

*Side: Be aware when we refer to cherry pick, we mean that the commit should be cherry picked and left as independent commit. While we rebase to get rid of own temporary commits and attempts, typically at the end of the process before merge, we don't get rid of history involving significance like if someone else has also contributed on parts of a combined work. Or if several topics is handled, then those are left as separate commits as well. For next time ;)*

andrerom:
Ok, seems good, true to look for additional data related to this object and could not find any, +1

clash82:
@andrerom thank you :) I didn't find anything more besides changes mentioned in this PR.

andrerom:
Thanks, merged :)

PR #180, opened 2015-09-05
Change Places to use findContentInfo api

andrerom:
Small optimization for places to not have to load full content object and throw it away again as code is only interested in main location id.


clash82:
+1

andrerom:
review ping @yannickroger

SylvainGuittard:
+1

yannickroger:
+1

PR #181, opened 2015-10-01
Fix EZP-24897: Error on blog post first draft preview

joaoinacio:
JIRA: https://jira.ez.no/browse/EZP-24897

Make sure not to attempt loading a `null` locationId when previewing a first draft.

(this has been resolved in master with the `@ParamConverter` in controllers)

### DIFF
--- a/Controller/DemoController.php
+++ b/Controller/DemoController.php
@@ -156,9 +156,20 @@ class DemoController extends Controller
      */
     public function showBlogPostAction( $locationId, $viewType, $layout = false, array $params = array() )
     {
-        // We need the author, whatever the view type is.
         $repository = $this->getRepository();
-        $location = $repository->getLocationService()->loadLocation( $locationId );
+        if ( isset( $params['location'] ) && $params['location'] instanceof Location )
+        {
+            $location = $params['location'];
+        }
+        else if ( is_numeric( $locationId ) )
+        {
+            $location = $repository->getLocationService()->loadLocation( $locationId );
+        }
+        else
+        {
+            throw new NotFoundHttpException( "Unknown location '$locationId' cannot be displayed." );
+        }
+        // We need the author, whatever the view type is.
         $author = $repository->getUserService()->loadUser( $location->getContentInfo()->ownerId );
 
         // TODO once the keyword service is available, load the number of keyword for each keyword


yannickroger:
+1
ping @lolautruche @bdunogier

andrerom:
please rebase with more recent master, this has parent from march..

Controller/DemoController.php
```diff
@@ -156,9 +156,17 @@ public function listBlogPostsAction( $locationId )
      */
     public function showBlogPostAction( $locationId, $viewType, $layout = false, array $params = array() )
     {
-        // We need the author, whatever the view type is.
         $repository = $this->getRepository();
-        $location = $repository->getLocationService()->loadLocation( $locationId );
+        // When previewing first draft there is no location id
+        if ( $locationId === null && isset( $params['isPreview'] ) && $params['isPreview'] == true && isset( $params['location'] ) )
```
bdunogier: We could use what is at `$params['location']`, if it is a valid location, instead of loading at all.

bdunogier:
Overall it won't hurt. It is a proper workaround.

Another thing you could do is replace `$locationId` with `Location $location`. The controller resolver will use `$params['location']`, and you won't have to care about it. It is actually the easiest workaround for this issue (it doesn't fix everything, but it makes custom controllers usable, as long as you don't rely on the locationId at all).

joaoinacio:
@andrerom, @bdunogier: do note this is a PR agains __5.4__ branch, not master (could have mentioned this explicitly in description). The CS is being maintained.
In master, as I mentioned, using a `Location $location` object works without any issue.

bdunogier:
Are you sure that typehinting doesn't work on 5.4 ?

andrerom:
> @andrerom, @bdunogier: do note this is a PR agains 5.4 branch, not master (could have mentioned this more explicitly in description). The CS is being maintained.

ok, then this is more clear indeed :)
Then maybe prefer to use location if set, and try to load if not, or add exception if both is not set also.

bdunogier:
On second thought, preview *would* work on 5.4 with typehinting. But view wouldn't anymore.

Reason: the [PreviewController](https://github.com/ezsystems/ezpublish-kernel-ee/blob/5.4/eZ/Publish/Core/MVC/Symfony/Controller/Content/PreviewController.php#L142) sets the fake location as the `location` request attribute, and the [`ControllerResolver`](https://github.com/symfony/http-kernel/blob/2.7/Controller/ControllerResolver.php#L108) will search for request attributes named as each controller action's parameters.

On master, it also works for view, since the `ParamConverter` does it for this part.

All of this to say that if we had the ParamConverter here, we would have a friendly workaround :-)

lolautruche:
+1

PR #182, opened 2015-10-04
Improved siteaccess configuration

bdunogier:
> Required by ezsystems/ezpublish-kernel#1425

Overall, defining the siteaccessess should be done at app level, not site bundle.

- Removed siteaccess groups config from `ezdemo.yml`
- Added siteacces config to install config `template.yml`

### DIFF
--- a/Resources/installer/config_templates/ezpublish.yml
+++ b/Resources/installer/config_templates/ezpublish.yml
@@ -21,10 +21,11 @@ ezpublish:
         # Available siteaccesses
         list:
             - demo_site
+            - fre
 
         # Siteaccess groups. Use them to group common settings.
         groups:
-            demo_site_group: [demo_site]
+            demo_site_group: [demo_site, fre]
         default_siteaccess: demo_site
         match:
             URIElement: 1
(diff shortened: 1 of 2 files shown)


Controller/DemoController.php
```diff
@@ -47,10 +47,10 @@ public function userLinksAction()
      * @param array $params
      * @return \Symfony\Component\HttpFoundation\Response
      */
-    public function showArticleAction($locationId, $viewType, $layout = false, array $params = array())
+    public function showArticleAction(Location $location, $viewType, $layout = false, array $params = array())
```
andrerom: is this part of another commit maybe? :)

Resources/installer/config_templates/ezpublish.yml
```diff
@@ -35,6 +36,7 @@ ezpublish:
     # System settings, grouped by siteaccess and/or siteaccess group
     system:
         demo_site_group:
+            translation_siteaccesses: [fre, eng]
```
andrerom: `eng` is not configured above, should it be added?

andrerom:
besides comments above, +1

Controller/DemoController.php
bdunogier: Shouldn't have been there, I'll remote it. Thank you.

Resources/installer/config_templates/ezpublish.yml
bdunogier: Should actually be `demo_site`, changing.

PR #184, opened 2015-10-07
EZP-24940: Remove unused Repsonse object in DemoBundle FolderController

wizhippo:
The response is create but not used.  Also the content is retrieved but not used.


andrerom:
Might rather be a bug? Or is the cache rules set somewhere else?

wizhippo:
the cache is handled by the ez_content controller

andrerom:
Ok clear, +1

ping @yannickroger @clash82 ?

wizhippo:
https://jira.ez.no/browse/EZP-24940

clash82:
Good work, thank you +1

PR #185, opened 2015-10-08
EZP-24938: Fix language switcher when no translation exists

wizhippo:
When a translation does not exist an exception is thrown.  This does not add the language if a valid url cannot be generated.


yannickroger:
You should create a Jira issue. We need it to process any type of fixes (except for typos)

wizhippo:
@yannickroger done thanks

yannickroger:
ping @clash82

clash82:
I would suggest to change commit name to include Jira ID, but besides this +1

andrerom:
+1 ping @lolautruche

PR #186, opened 2015-10-10
Applied the view controller refactoring

bdunogier:
This applies the changes from ezsystems/ezpublish-kernel#1454.

Content and Location controller actions now receive a `ContentView` object as their argument instead of individual parameters. The Content and Location objects are pre-loaded.

The actions can now return the View object, modified according to the needs of the controller, instead of building the Response.

### DIFF
--- a/Controller/DemoController.php
+++ b/Controller/DemoController.php
@@ -10,11 +10,12 @@ namespace EzSystems\DemoBundle\Controller;
 
 use eZ\Bundle\EzPublishCoreBundle\Controller;
 use eZ\Publish\API\Repository\Values\Content\Location;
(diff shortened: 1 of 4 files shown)


bdunogier:
The issue is (as far as I'm concerned) a known one. The login screen should show up on access denied errors. It does not. And I don't know (yet) why.

andrerom:
Reproduced on own install?

bdunogier:
Yes, @andrerom, reproduced.

bdunogier:
And login issue fixed in https://github.com/ezsystems/ezpublish-kernel/pull/1468.

PR #188, opened 2015-10-15
Implement EZP-24925: Extract XmlText field type out of kernel into separate repository

pspanja:
This PR resolves `DemoBundle` part of https://jira.ez.no/browse/EZP-24925

Corresponding PR on ezpublish-kernel: https://github.com/ezsystems/ezpublish-kernel/pull/1466

This converts XmlText fields to RichText fields.
Behat and REST tests in `ezpublish-kernel` depend on these fixtures.


andrerom:
+1

Side, what is `version=\"5.0-variant ezpublish-1.0\"` ?

pspanja:
That is required by DocBook, see 2.1 at http://www.docbook.org/tdg5/en/html/ch05.html

> If you change the DocBook schema, the namespace remains the same, but you must provide an alternate version identifier for the schema and the modules you changed. The version attribute identifies the version of DocBook the alternate is based on, specifies what type of variant it is, and > names the variant and any additional modules. While the format for the version string is not part of the normative specification, the DocBook Technical Committee recommends the following format:
>
> base_version-(subset|extension|variant) (name[-version])+
>
> For example, version 1.0 of Acme Corporation’s extension of DocBook V5.0 could be identified as “5.0-extension acme-1.0”.

So basically it says this is 1.0 version  of eZ Publish's DocBook 5.0 variant.
But now that you mention it, probably should have used "ezsystems" instead of "ezpublish".

andrerom:
Or ezplatform :)

andrerom:
And not 5.0 :)

Possible?

pspanja:
Sure it is possible, there is just a truckload of places to update :)
But 5.0 refers to the DocBook version used, so that should stay.

Then I would go for `5.0-variant ezplatform-1.0`.
BTW, right now only the presence of the version attribute is required by the schema, maybe I should also be strict about accepting only exact version. Actually I think we should do that as a rule, it would probably prevent all kind of issues in the future when the format is expanded/changed.

pspanja:
> Actually I think we should do that as a rule, it would probably prevent all kind of issues in the future when the format is expanded/changed.

I meant we should also dispatch conversion on version information.

andrerom:
I got that, yes why not.

pspanja:
I'll create new issue for that, this current one already expanded quite a bit.

pspanja:
Followups created:
* https://jira.ez.no/browse/EZP-24976 validate version attribute and dispatch conversion on it
* https://jira.ez.no/browse/EZP-24977 change the format of the version attribute

andrerom:
+1

PR #189, opened 2015-10-17
Alternative QueryController approach

bdunogier:
> Applies the concepts from https://github.com/ezsystems/ezpublish-kernel/pull/1473

Updates the blog view to use the QueryController.


clash82:
Awesome! can't wait to see this in action with pagination support :)

andrerom:
Any POV on how we should deal with Paging? Got feedback that PagerFanta is not really good compared to what we had in legacy, which was purly template based.

crevillo:
@andrerom Probably PagerFanta can require more work if you want to theme your paginations, but other than that, i'm really happy with it :)

bdunogier:
I don't think we can really skip some sort of support for a Pager here.

The fact that PagerFanta might not be right for everybody can be an issue. One way would be to make that extensible *to some extent*. It shouldn't be very complicated: the code that builds the actual result variable based on the search results can be refactored into an external service. We must support both pager & not pager (as it is, semantically, a QueryController option), and we should make sure that the pager implementation can easily be swapped with another one, by means of service container aliasing.

PR #191, opened 2015-11-23
EZP-25143: Changed installer to use ezplatform.yml

clash82:
Link to Jira task: https://jira.ez.no/browse/EZP-25143

This patch requires https://github.com/ezsystems/ezpublish-kernel/pull/1511


clash82:
ping @andrerom @bdunogier

andrerom:
Instead of overriding it would be better if it was extended, mainly reducing the amount of config here to avoid issues down the line.

@bdunogier anyway we can do that?
What about the following, given:
1. we need to separate config generation from db insert command *(for platform.sh and any kind of proper container setup)*
2. given ezpublish_prod.yml and ezpublish_dev.yml is empty anyway

Add a Composer command that generates these two on install, and then we let studio override this config generation command in it's compser.json, or this file can just be overridden / appended to as it is no longer in the repo anyway *(so no conflicts on upgrade: win!)*.

clash82:
ping @bdunogier

bdunogier:
I'd change the commit message so that it is meaningful from the perspective of this particular bundle. Could be like:

`EZP-25143: changed installer to use ezplatform.yml`

That aside, the code is fine, +1.

bdunogier:
What do you mean with the `ezplatform_dev.yml` and `ezplatform_prod.yml`  ? What would you put there ?

About config generation in general, I agree with the need to remove it from install. And actually, it is not required anymore, since there is an `ezplatform.yml` file by default. The install script shouldn't have to care for this anymore.

As I see it, any customization to `ezplatform.yml` for the demo should just be part of the demo's repository (e.g. fork of ezplatform).

Do you think we can work with that ?

andrerom:
> As I see it, any customization to ezplatform.yml for the demo should just be part of the demo's repository (e.g. fork of ezplatform).
> 
> Do you think we can work with that ?

ok, as my approach would require changes on root repo anyway, this would be cleaner indeed.

clash82:
So, can we forward with this? :)

andrerom:
+1 for now, the config cleanup we talked about is pending further work on install process.

andrerom:
Ideally should have re started travis to see how this bahaves now, so will need to track the merge commit instead: https://travis-ci.org/ezsystems/DemoBundle/builds/93146454 To know if there is any more followup work needed on this.

clash82:
Great, thanks.

PR #192, opened 2016-02-20
EZP-25485 - eZ Demo with video block but no content will return error 500

pedroresende:
# EZP-25485 - eZ Demo with video block but no content will return error 500

If you add a video block to a landing page object, but don't add any content to the video block, when accessing the landing page it will return a

Link: https://jira.ez.no/browse/EZP-25485

````
Key "0" does not exist as the array is empty in eZDemoBundle:block:video.html.twig at line 11
````

### DIFF
--- a/Resources/views/block/video.html.twig
+++ b/Resources/views/block/video.html.twig
@@ -4,14 +4,16 @@
         <h2>{{ block.name }}</h2>
     </div>
     {% endif %}
-    {{ render(
-        controller(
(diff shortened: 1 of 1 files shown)


andrerom:
+1

crevillo:
+1

yannickroger:
+1

yannickroger:
Thanks @pedroresende

PR #193, opened 2017-04-27
EZEE-1483: Limit Twig to 1.x due to invalid constraint in KnpMenuBundle 1.x which breaks with Twig 2.x

webhdx:
**JIRA: https://jira.ez.no/browse/EZEE-1483**

This PR has no real impact as DemoBundle is no longer maintained. Although DemoBundle is affected by invalid version constraint for `twig/twig` in `knplabs/knp-menu-bundle` package. This fix forces Twig 1.x to install.


webhdx:
@andrerom Fixed.

